PR #10, opened 2026-06-14
docs(spec): define app service boundary

Yevanchen:
## Summary

- Define App-local Service as the Project/App capability and resource unit.
- Reframe Agent as an Agent Service backed by the existing Agent identity, without introducing a second Agent table or App runtime.
- Add Project Templates as the path for simple one-Agent, one-Channel Apps and reject required App/Agent type pickers or type-driven persistence limits.
- Sync README, SPEC, architecture, PRD index, and Project/App Boundary PRD.

## Verification

- `bash ./init.sh development`
- `just fmt-check-path README.md`
- `just fmt-check-path docs/SPEC.md`
- `just fmt-check-path docs/prd/project-app-boundary.md`
- `just fmt-check-path docs/prd/README.md`
- `just fmt-check-path docs/architecture.md`
- `just check`
- `just commit-check`

Generated files: none.
GraphQL codegen: checked via `just check`; no generated output committed.
DB baseline/migrations: not changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,10 @@ For implementation language and V1 boundaries, read [SPEC.md](./docs/SPEC.md) be
 ## Current Product Lock
 
 - The primary loop is `PRD.md + @mosoo` -> create or select an App -> provision Agents and app-local resources -> expose an Agent through API or channel when needed -> export the App as `Skill.md`.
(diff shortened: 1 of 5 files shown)


samzong (approved):
/approve
/lgtm

PR #206, opened 2026-07-02
fix(api): harden keyless bound endpoint and guard prod D1 deploys

Yevanchen:
## Summary

Two evidence-backed, root-cause fixes surfaced by a full code audit. Both are the "truly severe → fix at root + prevent recurrence" class (leaked-URL billing/DoS, and silent prod schema divergence); the rest of the audit findings are scale-time optimizations we're deliberately deferring.

- **`fix(public-api)`** — the keyless, internet-facing `POST /api/v1/bound/:token` endpoint bypassed the rate limit and body-size cap every PAT-authenticated public route enforces. A single leaked capability URL (it lives in the URL path, is unrevocable, 10-year TTL) could be replayed to launch unbounded agent runs billed to the app owner and hold ~25s Worker slots each — a single-tenant cost-exhaustion / DoS. Fix reuses the existing limiter keyed on the capability identity (`bound:appId:agentId` bucket) and the shared streaming body-size cap. No new system added; the special-case bypass is removed so the route matches its siblings.
- **`fix(deploy)`** — `wrangler` records applied D1 migrations by **filename**, so the repo's rewrite-the-baseline workflow makes a new `CREATE TABLE` in `0000_baseline.sql` a silent no-op on a DB that already recorded the baseline; the Worker then deploys referencing tables that don't exist in prod. Adds a fail-fast guard between migration-apply and Worker-deploy that diffs the baseline's declared tables against the live prod schema and aborts on drift.

## Why

- Bound endpoint: leaked/replayed capability URL → unbounded owner-billed runs + Worker DoS. Directly threatens billing and availability for a single tenant.
- Deploy guard: the next schema-bearing change is near-certain to hit the silent-drop path (baseline has been rewritten 14× historically), producing a prod Worker whose queries reference missing tables — an invisible-until-runtime outage. Preserves move-fast baseline-rewrite while making that outcome impossible to ship unnoticed.

## Verification

- Commands: `bun test tests` in `apps/api` → **771 pass / 0 fail** (incl. 2 new test files, 10 new cases); `@mosoo/api tc` (typecheck) → exit 0; `vp lint` on touched files → clean; `vp fmt --check` → clean.
- New tests: `tests/prod-schema-guard.test.ts` (parse/diff/extract, incl. fail-closed on garbage output), `tests/bound-agent-body-limit.test.ts` (oversized body → `PublicApiError`, in-cap body parses).

## Risk And Blast Radius

- Affected packages: `@mosoo/api` only.
- Affected flows: bound-agent single-call API (now rate-limited at 120/min per capability + body cap), and the prod deploy script (adds a read-only `d1 execute` schema check before deploy).
- Rollback: revert; no data or schema migration involved.

## Reviewer Focus

- Rate-limit bucket key namespace (`bound:` prefix) not colliding with PAT tokenIds.
- The guard is **table-level** only (catches a missing table, the catastrophic case — not an added column on an existing table); noted in code comment. Runs after `applyD1Migrations`, so the correct incremental path always passes.
- Known deferrals (audit findings intentionally NOT in this PR): full per-account spend budget, IP-based limiting on the bound path, `channel-final-delivery-dlq` consumer, DB scale indexes, web bundle trims. These are scale-time optimizations, not launch blockers.

## Compatibility

- Public contract changes: none (adds 429/400 responses to a path that had no limits).
- Deployment: the deploy script now performs a read-only prod schema check; requires the same wrangler auth it already uses.

## Required Checks

- [x] PR title `type(scope): subject`
- [x] Commits `type(scope): subject`, English only
- [x] Diff scoped; no unrelated cleanup
- [x] UI/UX: N/A (backend only)
- [x] Generated files / DB baseline: none hand-edited

### DIFF
--- a/apps/api/bin/deploy-prod.ts
+++ b/apps/api/bin/deploy-prod.ts
@@ -1,5 +1,10 @@
 #!/usr/bin/env bun
 import type { BunRuntime } from "../../../config/bun-script-types";
+import {
+  extractTableNames,
+  findMissingProdTables,
+  parseExpectedTableNames,
+} from "./prod-schema-guard";
 
 declare const Bun: BunRuntime;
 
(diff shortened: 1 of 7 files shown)


Yevanchen:
Autopilot CI check: the failing run is post-merge on main, not a PR check. PR checks for this branch passed before merge, but the main workflow Notify docs OpenAPI sync failed here: https://github.com/langgenius/mosoo/actions/runs/28582155518. The failing step attempted repository_dispatch to langgenius/mosoo-docs and got HTTP 403 while MOSOO_DOCS_DISPATCH_TOKEN was present, so this looks like a token/repository permission issue rather than a branch code failure. I did not push a branch fix because the original PR is already merged and the failure needs secret/permission correction or a follow-up workflow change.

PR #207, opened 2026-07-02
fix(skills): persist large skill snapshot entries

RockChinQ:
## Summary

- Chunk skill snapshot entry inserts so larger skill archives stay under D1's per-statement bind limit.
- Repair existing content-addressed snapshots whose metadata row exists but entry rows were never fully persisted.
- Add regression coverage for large wrapped skill archives.

## Root Cause

`publishSkillSnapshot` inserted every `skill_snapshot_entry` row for a normalized skill package in one statement. Larger packages, such as the xlsx skill zip, exceeded the D1 bind limit during `/api/skill/package`, while `/api/skill/inspect` still worked because inspection does not persist entries.

## Validation

- `just test-file apps/api/tests/skill-package-snapshot.test.ts`
- `just test-file apps/api/tests/skill-package-source.test.ts`
- `just fmt-check-path apps/api/src/modules/skills/application/skill-package-snapshot.service.ts`
- `just fmt-check-path apps/api/tests/skill-package-snapshot.test.ts`
- Local HTTP upload to `/api/skill/package` returned 200 for `xlsx.zip`.
- Browser upload through `/integrations/skills` succeeded locally.

## Screenshot

Before:

image

After:

image

### DIFF
--- a/apps/api/src/modules/skills/application/skill-package-snapshot.service.ts
+++ b/apps/api/src/modules/skills/application/skill-package-snapshot.service.ts
@@ -7,12 +7,14 @@ import type { NormalizedSkillPackage, SkillPackageEntry } from "@mosoo/skill-pac
 import { and, asc, eq, inArray } from "drizzle-orm";
 
 import type { ApiBindings } from "../../../platform/cloudflare/worker-types";
-import { getAppDatabase } from "../../../platform/db/drizzle";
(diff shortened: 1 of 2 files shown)


RockChinQ:

RockChinQ:
Superseded by #208 using a ship-policy-compliant branch name.

PR #223, opened 2026-07-03
perf(web): tree-shake graphql documents out of the entry chunk

Yevanchen:
## Summary

- Add a build-time `graphql-document-optimizer` Vite plugin in `apps/web/vite.config.ts` that replaces each `graphql(/* GraphQL */ \`...\`)` call with the fully resolved document text from the generated artifact (`src/gql/graphql.ts`), so no chunk imports `src/gql` at runtime.
- Annotate generated `TypedDocumentString` documents with `/*#__PURE__*/` via the codegen `beforeOneFileWrite` hook (`config/graphql-codegen.ts`), and align `scripts/check-graphql-codegen.ts` with the combined hook.

## Why

Every `graphql()` call resolves through the generated `src/gql/gql.ts` runtime lookup map, which references **every** operation in the project. The map plus the full document set (~76 KB minified) was baked into the entry chunk and downloaded on every first page load, including the public `/login` route. The generated file itself warns about this and recommends the client preset's `babelOptimizerPlugin`, but `@vitejs/plugin-react` v6 is oxc-based and no longer runs Babel — and even the official import-rewrite is not enough here, because all documents live in one generated module that every route uses, so the whole module survives tree-shaking and stays on the critical path. Inlining the resolved text (operation + fragment dependencies — exactly the string `requestGraphQL` sends over the wire today) removes the runtime dependency on `src/gql` entirely; each chunk now carries only the documents it executes.

Production-build measurement, gzipped, identical methodology on both builds:

| Metric | Before | After | Δ |
| --- | --- | --- | --- |
| Entry chunk (`index-*.js`) | 42,509 B | 33,110 B | **−9,399 B (−22.1%)** |
| Entry static import closure (7 chunks, downloaded before the SPA boots) | 129,216 B | 119,817 B | **−9,399 B (−7.3%)** |
| Total emitted JS | 734,960 B / 96 files | 725,573 B / 96 files | −9,387 B (−1.3%) |

Total JS goes down (not just moves): the redundant lookup map and the per-page copies of unused documents are gone. `TypedDocumentString`, `gql.ts`, and `graphql.ts` no longer appear anywhere in the production bundle; each document string sits only in the chunk(s) that execute it (verified per-chunk with `grep` over `dist/assets`).

## Verification

- Commands: `vp exec tsc --noEmit` ✅ · `vp lint apps/web/vite.config.ts config/graphql-codegen.ts scripts/check-graphql-codegen.ts` ✅ · `bun run test` in `apps/web` (113 pass) ✅ · `bun run graphql:codegen:check` ✅ · `vp build` ✅
- Manual steps: served the production `dist/` with `vite preview` and loaded `/login` in headless Chromium — page renders identically (screenshot-verified), no page errors, and the boot-path `Viewer` GraphQL request is issued with the inlined document (the only console entry is the expected 502 from the absent local API). DOMContentLoaded 45 ms on the local preview.
- Not run: full e2e suite (no backend in this workspace).

## Impact

- User/API/contract changes: none — the wire format is byte-identical (the inlined text is exactly what `TypedDocumentString.toString()` returned via the map).
- Generated files / GraphQL / DB / lockfile: `apps/web/src/gql/graphql.ts` regenerated with `/*#__PURE__*/` annotations (deterministic; `graphql:codegen:check` passes).
- Risk and rollback: the transform is build-only (`apply: "build"`) — dev server, tests, and tsc keep the original call sites and the gql.ts map. If the artifact and a call site ever drift, the plugin fails the build with an actionable error instead of silently falling back to the map. Rollback: revert this commit.

## Review

- Closest review areas: the rewrite patterns in `apps/web/vite.config.ts` — they rely on the machine-enforced shape of codegen call sites (one definition per template, PascalCase names, no interpolation, no backticks), which is validated across all 101 call sites in this repo.
- Known trade-offs: a document shared by several lazy chunks is duplicated into each (small, compresses well; total JS still shrank 9.4 KB gzip). Build-vs-dev module graphs differ slightly for files containing `graphql()` calls (same trade-off the official babel optimizer makes).

### DIFF
--- a/apps/web/src/gql/graphql.ts
+++ b/apps/web/src/gql/graphql.ts
@@ -1329,7 +1329,7 @@ export class TypedDocumentString<TResult, TVariables>
     return this.value;
   }
 }
-export const LarkAgentChannelRegistrationFieldsFragmentDoc = new TypedDocumentString(`
+export const LarkAgentChannelRegistrationFieldsFragmentDoc = /*#__PURE__*/ new TypedDocumentString(`
     fragment LarkAgentChannelRegistrationFields on LarkAgentChannelRegistration {
   appId
   appSecret
@@ -1344,7 +1344,7 @@ export const LarkAgentChannelRegistrationFieldsFragmentDoc = new TypedDocumentSt
   userCode
 }
     `, {"fragmentName":"LarkAgentChannelRegistrationFields"}) as unknown as TypedDocumentString<LarkAgentChannelRegistrationFieldsFragment, unknown>;
-export const AgentChannelBindingFieldsFragmentDoc = new TypedDocumentString(`
+export const AgentChannelBindingFieldsFragmentDoc = /*#__PURE__*/ new TypedDocumentString(`
     fragment AgentChannelBindingFields on AgentChannelBinding {
   activityLastTriggeredAt
   activitySessionCount7d
@@ -1360,7 +1360,7 @@ export const AgentChannelBindingFieldsFragmentDoc = new TypedDocumentString(`
   updatedAt
 }
     `, {"fragmentName":"AgentChannelBindingFields"}) as unknown as TypedDocumentString<AgentChannelBindingFieldsFragment, unknown>;
-export const WeChatAgentChannelPairingFieldsFragmentDoc = new TypedDocumentString(`
+export const WeChatAgentChannelPairingFieldsFragmentDoc = /*#__PURE__*/ new TypedDocumentString(`
     fragment WeChatAgentChannelPairingFields on WeChatAgentChannelPairing {
   binding {
     ...AgentChannelBindingFields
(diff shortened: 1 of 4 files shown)


Yevanchen:
CI check note from the scheduled fixer:

The failing check was `PR Ship Policy`, not a code/test failure. The job failed because the head branch was `agent/evanmore-claude/1d06df75`, and the workflow blocks tool/agent-prefixed branches:

`Head branch 'agent/evanmore-claude/1d06df75' uses a blocked tool/agent prefix. Use type/scope-subject under feat|fix|docs|... instead.`

This cannot be fixed with a normal commit on the same PR branch because the PR head branch name itself violates the policy, and this PR is now closed. A replacement PR should be opened from a branch with an allowed prefix such as `perf/web-tree-shake-graphql-documents`.

PR #299, opened 2026-07-13
fix(api): retain commands after ambiguous Queue sends

### LINKED ISSUE #284: fix(api): retain commands when Queue send outcome is ambiguous
## Area

API

## Summary

`enqueueApiCommand` deletes the durable `api_command` row when `API_COMMAND_QUEUE.send()` throws. A queue send can be accepted remotely while the producer loses the response. When that happens, the delivered message has no ledger row to claim, is acknowledged, and is never executed.

This is a P1 reliability issue because session-run and App-deployment dispatch both depend on this command path. A session run can already have a persisted active Run and user message before dispatch is enqueued.

## Steps to reproduce

1. Use the local SQLite D1 fixture and an `API_COMMAND_QUEUE` stub.
2. Make `send()` retain the command message, then throw a simulated response-timeout error.
3. Call `enqueueApiCommand()` with a valid `scheduled_maintenance` payload.
4. Confirm that the command row has been deleted.
5. Deliver the retained message to `processApiCommandMessage()`.

## Expected behavior

The durable command record remains available for retry or reconciliation. A delivered message must not be acknowledged as completed merely because its ledger row is missing after an ambiguous producer-side send failure.

## Actual behavior

The producer deletes the command row after `send()` throws. The later-delivered message cannot be claimed and is acknowledged without executing the command.

## Evidence

- Producer deletion: `apps/api/src/modules/api-command/application/api-command-ledger.ts:95-101`
- Consumer acknowledges an unclaimable message: `apps/api/src/modules/api-command/application/api-command-processor.ts:535-545`
- Session Run persistence precedes dispatch enqueue: `apps/api/src/modules/runtime/application/session-runs/queue-run.service.ts:92-159`

Local fault injection reproduced the full sequence: the ledger was empty after the simulated timeout, and the retained message was acknowledged without processing. No remote Cloudflare resource was used. To avoid creating remote Cloudflare resources, requiring account credentials, or causing potential usage charges, this report uses local fault injection to validate the relevant failure boundary.

## Environment

Mosoo commit: `c14997f`
Local fault injection with Bun 1.3.14, SQLite D1 fixture, and a Queue stub.

## Contribution

I can provide more details.

## Checklist

AsperforMias:
## Summary

- Persist each API command with a `queue_delivery_pending` marker before attempting Queue delivery.
- Treat a successful durable write as accepted work: Queue send failures are recorded for scheduled redrive instead of failing the caller and changing the owning Run to terminal failure.
- Redrive both commands whose first send never began and commands whose producer response failed; duplicate deliveries remain safe through durable consumer claiming.
- Fixes #284.

## Why

A producer-side timeout can occur after Cloudflare Queue has accepted a message. The original path deleted the command record. The first revision retained that record but still threw to App deployment, which marked the DeploymentRun failed before a delivered or redriven command could execute. The durable outbox must be the acceptance boundary, not the synchronous Queue response.

## Verification

- Commands: `just test-file apps/api/tests/api-command-queue.test.ts` (6 passing); `just test-file apps/api/tests/app-deployment-service.test.ts` (19 passing); `just tc-package @mosoo/api`; `just commit-check`; `just check` (942 passing tests, plus format, docs, lint, type, and GraphQL checks).
- Manual steps: Used local SQLite D1 fixtures and Queue fault injection. Verified a retained-but-timeout message remains claimable, a definite Queue failure is redriven, a command persisted before its first send is redriven, and an App deployment remains `queued` instead of becoming `failed` when Queue delivery is deferred.
- Not run: A remote Cloudflare Queue end-to-end test. No Cloudflare account resources were available for an isolated remote deployment, so no remote resources were created and no account usage or billing was incurred. The local fault-injection regression covers the relevant outbox/consumer consistency boundary, but does not claim a remote Cloudflare incident was observed.

## Impact

- User/API/contract changes: An operation with a successfully persisted command now succeeds even when the immediate Queue send response fails; delivery becomes asynchronous and observable through the durable command record.
- Env or config changes: N/A.
- Risk and rollback: A caller can receive success before Queue delivery completes, and a recovery send can duplicate an accepted message. This is intentional at-least-once behavior; consumer claim CAS prevents duplicate command execution. Revert the two PR commits to restore the prior synchronous failure behavior.

## Review

- Closest review areas: API command ledger, scheduled maintenance redrive, App deployment dispatch.
- Known trade-offs: Scheduled redrive bounds recovery latency. The retained outbox marker prioritizes non-loss over immediate failure reporting.

### DIFF
--- a/apps/api/src/modules/api-command/application/api-command-ledger.ts
+++ b/apps/api/src/modules/api-command/application/api-command-ledger.ts
@@ -1,7 +1,7 @@
 import { apiCommandsTable } from "@mosoo/db";
 import type { ApiCommandId, ApiCommandKind, ApiCommandRow } from "@mosoo/db";
 import { createPlatformId } from "@mosoo/id";
-import { and, asc, eq, lt, or, sql } from "drizzle-orm";
+import { and, asc, eq, inArray, lt, or, sql } from "drizzle-orm";
 
 import { createErrorLogContext, logError } from "../../../platform/cloudflare/logger";
 import type { ApiBindings } from "../../../platform/cloudflare/worker-types";
@@ -14,6 +14,10 @@ export const API_COMMAND_LEASE_MS = 5 * 60 * 1000;
 export const API_COMMAND_LEASE_RENEWAL_INTERVAL_MS = API_COMMAND_LEASE_MS / 5;
 
 export const API_COMMAND_QUEUE_SEND_FAILED_CODE = "queue_send_failed";
+
+export const API_COMMAND_QUEUE_DELIVERY_PENDING_CODE = "queue_delivery_pending";
+
+const API_COMMAND_QUEUE_DELIVERY_PENDING_MESSAGE = "API command is awaiting queue delivery.";
 
 const API_COMMAND_QUEUE_SEND_FAILED_MESSAGE = "API command queue send failed.";
 
@@ -95,7 +99,10 @@ async function clearApiCommandQueueSendFailure(input: {
       and(
         eq(apiCommandsTable.id, input.commandId),
         eq(apiCommandsTable.status, "queued"),
-        eq(apiCommandsTable.lastErrorCode, API_COMMAND_QUEUE_SEND_FAILED_CODE),
+        inArray(apiCommandsTable.lastErrorCode, [
+          API_COMMAND_QUEUE_DELIVERY_PENDING_CODE,
+          API_COMMAND_QUEUE_SEND_FAILED_CODE,
+        ]),
       ),
     )
     .run();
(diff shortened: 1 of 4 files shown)


AsperforMias:

Yevanchen (approved):
Reviewed with full codebase context. Verified the outbox design end to end: consumer claim CAS + ack-on-claim-miss makes duplicate redrive sends safe; the marker-clear update is guarded by status='queued' plus a marker whitelist so it cannot clobber real retry errors written by releaseApiCommandForRetry; the only production reader of api_command.lastErrorCode (deployment recoverMissingDispatch) also requires attemptCount >= max, so the new pending marker cannot fail runs prematurely. Session-run latency is unaffected thanks to the inline waitUntil dispatch; with the 1-minute cron, redrive latency is bounded and the LIMIT 100 scan is covered by the (status, updated_at) index. Tests exercise the real scheduled handler and dispatch path. LGTM.

PR #301, opened 2026-07-13
fix(cost): retain daily rollups for 180 days

### LINKED ISSUE #286: fix(cost): retain daily usage rollups for both 90-day comparison windows
## Area

Database

## Summary

The cost PRD requires daily usage rollups to be retained for 180 days so both the current and previous 90-day windows remain queryable. The cleanup path retains only about 97 days: it first applies the seven-day detail cutoff, then deletes daily rows older than 90 days before that cutoff.

After the system has been running long enough, `LAST_90_DAYS` can have correct current totals while `previousTotals` is incomplete or zero. This makes the cost dashboard unsuitable for the comparison it exposes.

## Steps to reproduce

1. Set the rollup job time to `2026-07-13T12:00:00Z`.
2. Insert a daily rollup dated `2026-03-01`.
3. Apply the cleanup predicate used by `runUsageDailyRollup`: `date(cutoffDate, '-90 days')`, where `cutoffDate` is seven days before the job time.
4. Query `usage_daily_rollup`.

## Expected behavior

A daily rollup needed by the previous 90-day comparison window remains available. Retention must be at least 180 days, as specified by the cost PRD.

## Actual behavior

The `2026-03-01` rollup is deleted. It belongs to the previous 90-day comparison period but is older than the approximately 97-day implementation retention window.

## Evidence

- Cleanup implementation: `apps/api/src/modules/cost/application/cost-rollup.service.ts:17-21,101-103`
- Organization card calculates a previous period: `apps/api/src/modules/cost/application/cost-query.service.ts:110-142`
- Product contract requires 180-day rollup retention: `docs/prd/cost-dashboard.md:80-91`

Local deterministic time-window fault injection reproduced the deletion. No remote Cloudflare resource was used. To avoid creating remote Cloudflare resources, requiring account credentials, or causing potential usage charges, this report uses local fault injection to validate the relevant failure boundary.

## Environment

Mosoo commit: `c14997f`
Local SQLite D1 fixture with deterministic UTC timestamps.

## Contribution

I can provide more details.

## Checklist

AsperforMias:
## Summary

- Retain daily usage rollups for 180 days, covering both current and previous 90-day comparison windows.
- Extract the production retention cutoff and predicate so the retention boundary is directly testable.
- Fixes #286.

## Why

The prior cleanup used a 90-day window relative to an already seven-day-old detail cutoff, retaining roughly 97 days. That deletes data required by the cost dashboard previous-period comparison and can make historical totals incomplete.

## Verification

- Commands: `just test-file apps/api/tests/cost-rollup.test.ts` (1 passing); `just check` (939 passing tests, plus format, docs, lint, type, and GraphQL checks).
- Manual steps: Used deterministic UTC time with the local SQLite D1 fixture. Applied the exact production retention predicate, verified a date within the 180-day horizon remains, verified an older date is removed, and queried the real `LAST_90_DAYS` comparison path.
- Not run: A remote D1 retention-job execution. No Cloudflare account resources were available for an isolated remote deployment, so no remote resources were created and no account usage or billing was incurred. The deterministic local regression verifies the retention and read-query boundary; this is not a claim that a remote Cloudflare incident was observed.

## Impact

- User/API/contract changes: No public API changes; historical cost comparison data is retained according to the product requirement.
- Env or config changes: N/A.
- Risk and rollback: Daily rollups consume up to roughly twice the prior retention storage. Revert this commit to restore the previous retention cutoff, with the associated loss of previous-window history.

## Review

- Closest review areas: Cost rollup service, cost query service, daily-rollup retention policy.
- Known trade-offs: The additional 83 days of retained aggregate rows increase D1 storage modestly; aggregate rows are small and required for the documented 90-day comparison.


AsperforMias:

Yevanchen (approved):
Verified against the actual query semantics: getOrganizationBillingCostCard computes the previous LAST_90_DAYS window as resolveCostWindow(range, new Date(window.sinceMs - 1)), whose oldest needed rollup date is today-179, while the old cleanup predicate retained only ~97 days -- so previousTotals was silently truncated once the system ran long enough. The 180-day cutoff covers the previous window with a one-day margin, and the pre-streamline PRD explicitly documented the 180-day retention contract. The test pins the exact boundary (cutoff 2026-01-14, boundary row 2026-01-15 retained) through the real production predicate and real queryTotals path. LGTM.

PR #308, opened 2026-07-14
fix(channels): retain final-delivery jobs after Queue send failure

### LINKED ISSUE #303: fix(api): retain channel final-delivery jobs after ambiguous Queue sends
## Area

API

## Summary

Channel final-delivery jobs are written to D1 before CHANNEL_FINAL_DELIVERY_QUEUE.send(), but the producer deletes the job whenever send() rejects. A producer-side timeout or connection reset does not prove that Cloudflare Queue discarded the message: Queue may already have accepted it.

If that happens, the retained Queue message reaches the consumer after its D1 ledger row has been deleted. The consumer logs channel-final-delivery.ledger_missing and acknowledges the message without sending the final reply.

This is the same distributed-systems failure mode addressed for api_command in #284 / #299, but it remains in the separate channel_final_delivery_job pipeline.

## Steps to reproduce

1. Use the local SQLite D1 fixture and the channel final-delivery Queue stub.
2. Create a completed Telegram final-delivery job.
3. Replace CHANNEL_FINAL_DELIVERY_QUEUE.send() with a fault-injection implementation that first records the message as accepted, then throws "producer timeout after Queue acceptance".
4. Call enqueueChannelFinalDeliveryJob() for a second external event.
5. Observe that the producer rejects and the Queue still contains the target jobId.
6. Query channel_final_delivery_job for that jobId: the row has been deleted by the producer error path.
7. Deliver the retained Queue message to processChannelFinalDeliveryMessage().

## Expected behavior

A successfully persisted final-delivery job must remain durable when the immediate Queue producer outcome is ambiguous. It should be marked pending/failed for a safe redrive, and duplicate Queue messages must remain harmless through the existing job claim/idempotency logic.

The consumer must not silently discard a message solely because the producer previously removed its ledger after an ambiguous send response.

## Actual behavior

The local fault injection reproduces the loss sequence:

- the Queue stub recorded the target message;
- enqueueChannelFinalDeliveryJob() rejected after the simulated timeout;
- the target ledger row was absent before consumption;
- the consumer logged channel-final-delivery.ledger_missing and acknowledged the retained message.

No provider final-reply request was made.

## Evidence

Relevant code:

- Producer deletes the durable job in its Queue.send() catch path: apps/api/src/modules/channels/application/channel-final-delivery-jobs.ts:271-277.
- Consumer acknowledges a missing ledger: apps/api/src/modules/channels/application/channel-final-delivery.service.ts:129-137.
- #299 fixes the analogous API-command path but its diff only changes api-command files and scheduled redrive, not channel final delivery.

Local fault-injection result:

    {
      "ledgerBeforeConsumption": false,
      "consumerActions": [{ "type": "ack" }]
    }

## Production impact

Cloudflare Queue producer outcomes are not an atomic transaction with D1. A timeout after Queue acceptance is a normal distributed-systems ambiguity, not evidence that the message was rejected.

In production, this can permanently suppress a completed Agent's final reply in Slack, Telegram, Discord, Lark, or WeChat. The originating run may complete successfully while the user sees only the working response or no terminal response. The current consumer acknowledgement removes the last retriable signal, making the loss silent and difficult to reconcile. External webhook retries may also create duplicate upstream handling depending on the provider.

## Environment

Mosoo commit: e6401855d0b20a2b2bcf12d9a14659e443176d2b.

Local Bun 1.3.14, SQLite D1 fixture, and injected Queue stub. No remote Cloudflare Queue, provider credential, model-provider call, or billable Cloudflare resource was used. This demonstrates the local D1/Queue consistency boundary; it is not a claim that a remote production incident was observed.

## Contribution

I can submit a PR.

## Checklist

AsperforMias:
Closes #303.

## Summary

- Persist a durable final-delivery ledger row before attempting Queue dispatch.
- Retain the row when Queue `send()` rejects or reports an ambiguous failure after acceptance.
- Redrive retained rows from the Worker scheduled handler.

## Why

A final response could be lost when D1 insert succeeded but Queue dispatch failed: the previous path deleted the only durable delivery record.

## Verification

- Commands: `just test-file apps/api/tests/channel-final-delivery-scheduling.test.ts` (15 pass); `just tc-package @mosoo/api`.
- Manual steps: injected both an ambiguous Queue result (message accepted, then `send()` throws) and a rejected Queue send. The real final-delivery ledger, Worker scheduled handler, queue consumer, and in-memory D1 path retained the job, redrove it, delivered it, and ACKed it.
- Not run: `just check` was not run; no real Cloudflare production account or remote Queue/D1 was used.

## Impact

- User/API/contract changes: None.
- Risk and rollback: The scheduled redrive scans a bounded set of retained jobs. Revert this PR to stop new redrives; retained ledger rows remain available for controlled recovery.

## Review

- Closest review areas: channel final-delivery ledger, Worker scheduled handling, queue idempotency.
- Known trade-offs: Cloudflare platform behavior was not tested against a real account. The local verification uses the real D1/Worker/service chain with Queue behavior injected only at the Cloudflare binding boundary.

### DIFF
--- a/apps/api/src/modules/channels/application/channel-final-delivery-jobs.ts
+++ b/apps/api/src/modules/channels/application/channel-final-delivery-jobs.ts
@@ -2,8 +2,9 @@ import { channelFinalDeliveryJobsTable } from "@mosoo/db";
 import type { AgentChannelBindingProvider, ChannelFinalDeliveryJobId } from "@mosoo/db";
 import { createPlatformId } from "@mosoo/id";
 import type { ChannelBindingId, SessionId, SessionRunId } from "@mosoo/id";
-import { and, eq } from "drizzle-orm";
+import { and, asc, eq, inArray } from "drizzle-orm";
 
+import { createErrorLogContext, logError } from "../../../platform/cloudflare/logger";
 import type { ApiBindings } from "../../../platform/cloudflare/worker-types";
 import { getAppDatabase, getD1ChangeCount } from "../../../platform/db/drizzle";
 import { currentTimestampMs } from "../../../time";
@@ -56,6 +57,14 @@ export interface JobLedgerRow {
 }
 
 const DELIVERY_CLAIM_PREFIX = "delivery_claim:";
+
+export const CHANNEL_FINAL_DELIVERY_QUEUE_DELIVERY_PENDING_CODE =
+  "channel_final_delivery_queue_delivery_pending";
+
+export const CHANNEL_FINAL_DELIVERY_QUEUE_SEND_FAILED_CODE =
+  "channel_final_delivery_queue_send_failed";
+
+const CHANNEL_FINAL_DELIVERY_QUEUE_REDRIVE_LIMIT = 100;
 
 export interface ChannelFinalDeliveryClaim {
   attemptCount: number;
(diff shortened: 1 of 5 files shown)


Yevanchen (commented):
Reviewed the durable outbox/redrive flow and did not find a code-level blocker in this pass. This still should not merge now: Channels is deliberately unavailable as an end-to-end product surface (no reachable setup UI or provider live-account smoke), and #303 is deferred rather than fixed.

Keep this draft; when Channels is revived, rebase onto current main and rerun the current repository plus live-boundary verification.

PR #312, opened 2026-07-14
fix(api): recover public idempotency completion failures

### LINKED ISSUE #305: fix(api): retain public API idempotency after completion write failures
## Area

API

## Summary

The public API idempotency wrapper returns the successful business response even when persisting the completed replay record fails. The reservation remains in the processing state, then the normal 10-minute cleanup deletes it. A later request with the same `Idempotency-Key` is treated as a new operation.

For `POST /api/v1/agents/:agentId/threads`, the idempotency key is not propagated as a downstream Thread or Run uniqueness key. Once the reservation expires, the retry creates another Thread and can start another Agent Run.

## Steps to reproduce

1. Use the local SQLite D1 public Thread API fixture and the real public HTTP route.
2. Wrap `D1Database.prepare()` so the first `UPDATE public_api_idempotency_key ... SET response_json, response_status` throws after the Thread creation operation has completed.
3. Send `POST /api/v1/agents/:agentId/threads` with a valid access token and `Idempotency-Key: completion-failure-retry`.
4. Observe a `201` response. The route logs `public-api.idempotency_completion_failed`, but the Thread has already been created.
5. Set the resulting `public_api_idempotency_key.updated_at` to an old value in the local fixture to model the documented 10-minute processing TTL.
6. Send the exact same request with the same idempotency key.
7. Query `session` for the target Agent.

## Expected behavior

Once a public mutation has succeeded, retries with the same idempotency key must not execute the business operation again, even if persisting the replay payload was interrupted.

The completion state should be atomically durable with the business mutation, or the reservation must remain recoverable and block/reconcile retries rather than being discarded after the processing TTL.

## Actual behavior

The local fault injection reproduced two successful non-replayed creates:

- the first request returned `201` after the injected completion-write failure;
- after expiring the incomplete reservation, the same request returned a second `201` with no `Idempotency-Replayed` header;
- the fixture contained two Thread rows for the Agent.

## Evidence

Relevant code:

- Successful completion persistence is caught and only logged: `apps/api/src/adapters/http/routes/public-api-route-support.ts:290-304`.
- Incomplete reservations are deleted after ten minutes: `apps/api/src/modules/public-api/public-api-idempotency.service.ts:233-254`.
- Thread creation generates a new session and passes `clientRequestId: null` when starting the initial Run: `apps/api/src/modules/public-api/public-thread-create.ts:67-129`.

This is not limited to Thread creation. The shared `runPublicApiIdempotentJson()` wrapper is used by public API mutations that opt into a body hash.

## Environment

Mosoo commit: `e6401855d0b20a2b2bcf12d9a14659e443176d2b`.

Local Bun 1.3.14, SQLite D1 public HTTP fixture, and an injected D1 completion-write failure. No remote Cloudflare account, Queue, provider credential, model-provider request, or billable Cloudflare resource was used. This demonstrates the local business-mutation/idempotency durability boundary; it is not a claim that a remote production incident was observed.

## Contribution

I can submit a PR.

## Checklist

AsperforMias:
Fixes #305

## Summary

- Keep incomplete `public_api_idempotency_key` reservations for the full existing 24-hour idempotency retention window. The former 10-minute processing cleanup no longer releases a key after the business mutation may have succeeded.
- Persist the public request's `Idempotency-Key` on the created Session metadata. On a stale retry of `POST /api/v1/agents/:agentId/threads`, locate the Session by Agent, PAT token, and key, then reconstruct and return the already-created Thread/Run instead of calling `createPublicThread()` again.
- For `POST /api/v1/threads/:threadId/events`, do not guess or fabricate a replay payload. A stale incomplete reservation remains retained and the exact retry receives `409 idempotency_conflict` with `Retry-After`; no second message, Run, queue submission, or billable Agent execution is created.
- Split implementation and regression coverage into separate commits.

## Why

#305 reproduced this failure sequence:

1. The public API inserts the unique `(token_id, idempotency_key)` processing reservation.
2. Thread creation commits the Session and initial Run.
3. The final `UPDATE public_api_idempotency_key SET response_json, response_status` fails.
4. The route returns `201`, but the reservation remains incomplete.
5. After ten minutes, the old cleanup deletes the reservation.
6. Retrying the identical request creates another Session/Thread and another Run.

The fix gives every stale incomplete reservation one safe outcome during the idempotency window: recover a durable Thread result when available, or retain and block it when a trustworthy generic replay result is unavailable.

## Verification

- Commands:
  - `just check`
  - `just test-file apps/api/tests/public-thread-api.e2e.test.ts`
- Manual / injected path:
  1. Use the local SQLite D1 public API fixture and the real Hono public HTTP route.
  2. Wrap `D1Database.prepare()` so the first final idempotency completion update throws only after the business mutation has succeeded.
  3. Submit the request with a valid PAT and a fixed `Idempotency-Key`.
  4. Age the incomplete reservation past 10 minutes and repeat the identical HTTP request.
- Assertions:
  - Thread creation: first request returns `201`; retry returns `201` with `Idempotency-Replayed: true`, the original Thread and Run IDs, and no second Thread.
  - Event submission: first request creates one message and one Run; retry returns `409 idempotency_conflict`, retains the reservation, and leaves message and Run counts unchanged.
  - Existing API E2E cases continue to cover normal key replay, mismatched request/key conflict, rate-limit cleanup, and requests without an idempotency key.
- Not run:
  - No remote Cloudflare D1/Worker/Queue deployment. No Cloudflare account, provider credential, model request, or billable resource was used. The local Worker/API route, D1 persistence, and D1 completion-write failure boundary were exercised directly.

## Impact

- User/API/contract changes:
  - Normal successful idempotency records keep their existing exact stored-response replay behavior.
  - In the exceptional completion-write-failure path, stale Thread creation retries reconcile to the already-created resource.
  - In the same exceptional event path, stale retries are safely rejected for the remainder of the 24-hour idempotency window rather than re-executed.
- Generated files / GraphQL / DB / lockfile: N/A. No migration, generated output, contract schema, configuration, or lockfile changes.
- Env or config changes: N/A.
- Risk and rollback:
  - The only deliberate availability trade-off is retaining an ambiguous event reservation until its existing 24-hour key window expires. This prevents a worse duplicate execution and billing outcome.
  - Reverting both commits restores the former 10-minute retry behavior and therefore reintroduces the duplicate-execution risk described in #305.

## Review

- Closest review areas:
  - `public-api-idempotency.service.ts`: reservation state, retention, and conflict retry time.
  - `public-api-route-support.ts`: stale reservation recovery versus safe rejection.
  - `public-thread-create.ts` and `public-thread-store.ts`: Session-backed recovery anchor and ownership predicate.
  - `public-thread-api.e2e.test.ts`: real route/D1 fault injection and count assertions.
- Known trade-offs:
  - Thread recovery reconstructs from the durable Session state after the original completion record was lost; it prevents a duplicate business operation but is not a byte-for-byte stored receipt.
  - Events are blocked instead of claiming a replay result when no durable receipt exists.


AsperforMias:
carefully review this

PR #317, opened 2026-07-14
fix(api): revoke bound deployment capabilities

### LINKED ISSUE #287: fix(api): revoke bound Agent capabilities when a deployment is deleted
## Area

API

## Summary

A deployed App receives a long-lived bearer capability URL for each bound Agent. The token contains only `appId`, `agentId`, `expose`, and expiry. Bound-call validation checks that the Agent remains published, but does not validate the Deployment, binding revision, or deletion state.

Deleting a deployment therefore does not revoke an already issued capability URL. If the URL leaks, it can continue to start owner-billed Agent Runs while the Agent remains published.

## Steps to reproduce

1. Mint a valid App Agent capability token for a published Agent with a live deployment version.
2. Simulate deletion of the external App deployment while leaving the Agent published.
3. Verify the token with `verifyBoundAgentCapability()`.
4. Apply `ensureBoundAgentServable()` to the published Agent.

## Expected behavior

Deleting a deployment or removing an Agent binding revokes all capabilities minted for that deployment or binding. The bound-call path must reject a revoked capability before starting a Run.

## Actual behavior

The token still verifies and the published-Agent guard still passes. The claims have no deployment or binding identity to revoke, and the guard only checks `appId`, publication status, and `liveDeploymentVersionId`.

## Evidence

- Ten-year capability TTL and token minting: `apps/api/src/modules/apps/application/app-deployment-executor.service.ts:749-787`
- Claims omit deployment and binding identity: `apps/api/src/modules/public-api/app-agent-capability.ts:16-21`
- Bound-call guard does not inspect deployment state: `apps/api/src/modules/public-api/app-agent-bound-call.ts:88-102`

Local cryptographic capability and published-Agent-guard fault injection confirmed that a valid token remains accepted for a published Agent and carries no deployment identifier. No model provider call or remote Cloudflare resource was used. To avoid creating remote Cloudflare resources, requiring account credentials, or causing potential usage charges, this report uses local fault injection to validate the relevant failure boundary.

## Environment

Mosoo commit: `c14997f`
Local Bun 1.3.14 fault injection.

## Contribution

I can provide more details.

## Checklist

AsperforMias:
## Summary

- Fixes #287.
- Add deployment, successful deployment-run, and public Agent binding provenance to minted bound-Agent capability tokens.
- Authorize that provenance against D1 before rate limiting, Session creation, Run enqueueing, or owner-billed execution begins.
- Repeat the deployment, revision, and Agent revocation predicate in the final `INSERT INTO session_run`, closing the deletion-versus-Run-creation race.
- Reject and classify expired, deleted, unpublished, binding-removed, and superseded-revision capabilities without logging capability URLs.

## Why

A signed capability previously proved only that a token had been minted for an App/Agent. It did not prove that the deployment and binding that delegated that authority were still current. After a deployment was deleted or a successful deployment removed/replaced a binding, a leaked URL could still start a new owner-billed Run.

The preflight D1 read alone left a time-of-check/time-of-use window: deletion could commit after the read but before the Run record was inserted. The final Run insertion now has the same D1 authority predicate, so the D1 statement linearizes revocation against billed work.

## Verification

- Commands:
  - `just test-package @mosoo/api` (977 pass, 0 fail)
  - `just test-file apps/api/tests/app-agent-capability-revocation-http.test.ts`
  - `just test-file apps/api/tests/app-agent-bound-run-revocation.test.ts`
  - `just test-file apps/api/tests/app-deployment-capability-authority.test.ts`
  - `just test-file apps/api/tests/app-agent-bound-call.test.ts`
  - `just test-file apps/api/tests/app-deployment-service.test.ts`
  - `just commit-check`
  - `git diff --check`
- Manual steps:
  - Exercised the actual bound HTTP route with local D1 and signed capability tokens. Deleted deployments and removed bindings returned before Session creation; the tests assert zero created Sessions.
  - Exercised the normal deployment delete service with an injected Cloudflare deletion client, then verified the persisted D1 authority state rejects the previously valid capability.
  - Injected a D1 deployment deletion immediately before the actual `INSERT INTO session_run` executes. The guarded queue path rejected the insert and asserted `session_run = 0`; the non-revoked control path created one queued Run.
  - Verified a failed newer deployment preserves the last successful capability, while a successful version rollover invalidates its predecessor.
- Not run:
  - No real Cloudflare account or remote Worker/D1 deployment was used. The Cloudflare boundary is injected locally; no remote CF resources, deployment side effects, or account charges occurred.
  - Root `just check` was attempted after the focused/full API checks but this machine intermittently reports an unrelated Web package-resolution error. The API typecheck and API suite pass; GitHub CI remains the repository-wide gate.

## Impact

- User/API/contract changes:
  - **BREAKING AND NECESSARY: capability URLs minted by older API versions lack deployment/binding provenance and are now rejected with `401`. They cannot be safely mapped to a revocation boundary; accepting them would retain the security and billing exposure fixed here. Re-deploy Apps that expose bound Agents to mint replacement URLs before enforcing this API version.**
  - **Changed authorization boundary: a bound URL now requires a live deployment, its latest successful matching deployment run, the claimed public binding, and a published matching Agent. The final Run insert repeats the revocation predicate atomically, so a revocation that commits during the request cannot create owner-billed work.**
  - **Operational effect: each accepted bound request adds D1 authority reads plus one bounded predicate on the existing Run insert. Rejections emit structured reason-only logs with IDs and binding metadata, never the capability URL/token.**
  - None. No schema, migration, generated GraphQL, or lockfile changes.
- Env or config changes:
  - None.
- Risk and rollback:
  - Rolling back to the prior API implementation restores acceptance of old URLs and the deletion/Run-creation race; use only for an emergency availability rollback.
  - Direct deletion of CF resources outside Mosoo remains a separate control-plane drift concern. A follow-up should reconcile external CF state with D1; this PR does not add a CF API call to every billed request.

## Review

- Closest review areas:
  - `apps/api/src/modules/public-api/app-agent-bound-ask.service.ts`
  - `apps/api/src/modules/apps/application/app-deployment-capability-authority.service.ts`
  - `apps/api/src/modules/runtime/infrastructure/session-runs/session-run-write.repository.ts`
- Known trade-offs:
  - Legacy capabilities fail closed rather than receive an unsafe compatibility grace period.
  - Authorization is driven by Mosoo persisted control-plane state. Out-of-band CF console changes require separately scheduled reconciliation rather than a per-request CF dependency.


AsperforMias:
it's necessary to review this **carefully and carefully** (cuz i dont have a relevant cf account to test prod online/offline)

**it's best to test this changes in ur cf platform to ensure**

and, must look these: (**IMPORTANT**)

>   * **BREAKING AND NECESSARY: capability URLs minted by older API versions lack deployment/binding provenance and are now rejected with `401`. They cannot be safely mapped to a revocation boundary; accepting them would retain the security and billing exposure fixed here. Re-deploy Apps that expose bound Agents to mint replacement URLs before enforcing this API version.**

Issue #322, opened 2026-07-14
fix(cost): make model-call and usage-ledger persistence recoverable


AsperforMias:
## Area

API

## Summary

**Severity: P2 - cost ledger atomicity and recoverability.**

The runtime persistence path writes the model-call record and the cost ledger record as two separate D1 operations. `upsertSessionModelCallUsage()` first commits `session_model_call`, then awaits `recordRuntimeUsageEvent()` to write `usage_event`.

If the second write fails after the first has committed, Mosoo durably records that a model call occurred (including its reported/token-derived cost) while omitting the same consumption from the `usage_event` ledger. The request rejects, so an immediate runtime retry can repair it; however, recovery currently depends on receiving an equivalent retry. There is no atomic D1 boundary, durable outbox, or reconciliation path that finds `session_model_call` rows missing their ledger counterpart.

This is a serious financial-data consistency problem. It does not assert that a provider or Cloudflare account was incorrectly charged. Rather, Mosoo can retain contradictory internal records and understate consumption in App Usage, exports, and daily rollups after a transient or ambiguous D1 write failure. The missing amount can accumulate without an operator-visible reconciliation signal. If the ledger becomes an input to quotas, credits, chargeback, or customer billing, this must be treated as P1.

## Steps to reproduce

1. Create the local SQLite D1 fixture containing a valid App, Agent, Session, Session Run, `session_model_call`, and `usage_event` schema.
2. Call the real `upsertSessionModelCallUsage()` service with a completed Run and a valid usage payload containing a stable `callId`, non-zero input/output tokens, and a declared usage contract.
3. Wrap only the D1 statement used by `INSERT INTO usage_event ... ON CONFLICT ...` so its first execution throws after the earlier `session_model_call` upsert has been allowed to commit.
4. Observe that the service rejects with the injected ledger-write failure.
5. Query both tables before any replay:

   ```sql
   SELECT COUNT(*) FROM session_model_call;
   SELECT COUNT(*) FROM usage_event;
   ```

## Expected behavior

A model-call fact and its corresponding usage-ledger fact must not be left in contradictory durable states.

The implementation should either:

- commit the model-call update and ledger update in one atomic D1 business boundary; or
- persist a durable, idempotent recovery obligation and reconcile it until the ledger fact exists exactly once.

A retry after an ambiguous failure must converge to one `session_model_call` and one matching `usage_event`, without duplicate cost. Failure and recovery must be covered by an injected D1 test.

## Actual behavior

[...]

PR #324, opened 2026-07-14
fix(cost): retain reported usage without tokens

### LINKED ISSUE #321: fix(cost): retain reported USD usage when token counters are absent
## Area

API

## Summary

**Severity: P2 - cost ledger correctness.**

`SessionUsageSummary` permits a runtime to report a valid USD `costAmount` while every token counter is absent. This is a valid shape for an upstream/aggregator that provides billed cost but no token breakdown. However, `recordRuntimeUsageEvent()` returns before it validates or persists the reported USD amount when all four token buckets normalize to zero.

The result is an internal consumption inconsistency: `session_model_call` can retain the reported cost, but the canonical `usage_event` ledger receives no row. App Usage, CSV export, seven-day detail queries, and later daily rollups all omit that consumption.

The Cost PRD explicitly says that an unknown model may retain a reported USD amount. The current early return prevents exactly that fallback when token fields are unavailable.

This is not a claim that a remote provider or Cloudflare account was overcharged, and the current Cost dashboard is not a provider invoice. It is nevertheless serious: the product's own spend signal can deterministically understate actual reported model consumption. The undercount is unbounded across affected requests and can mislead operational spend decisions. If `usage_event` later feeds budgets, quota enforcement, chargeback, or billing, this boundary becomes P1.

## Steps to reproduce

1. Use the local SQLite D1 `usage_event` fixture and invoke the real `recordRuntimeUsageEvent()` service.
2. Supply a valid persisted Run context and this valid normalized usage shape:

   ```ts
   {
     costAmount: 0.42,
     costCurrency: "USD",
     source: "session_update",
     usageContract: "openai_total_with_cached_breakdown",
   }
   ```

   Do not supply `inputTokens`, `outputTokens`, `cachedReadTokens`, or `cachedWriteTokens`.
3. Query `usage_event` after the service resolves.
4. Optionally inspect `session_model_call` through the surrounding model-call persistence path: it has a field for the same reported USD cost, so the two persistence surfaces can disagree.

## Expected behavior

A valid non-negative reported USD cost must be represented in the usage ledger even when token counters are unavailable. For an unknown model, the persisted row should retain the provided USD amount, use the existing unknown-pricing semantics, and keep zero token buckets.

The fix must preserve the existing `(source, source_event_id)` upsert behavior so retries update one logical usage fact rather than add a second cost row.

## Actual behavior

`recordRuntimeUsageEvent()` normalizes missing token fields to zero and returns when all four token buckets are zero. It therefore returns before `toProvidedUsdCost()` and `calculateUsageCost()` can retain the `$0.42` fallback.

The local D1 fault injection deterministically observed:

- `recordRuntimeUsageEvent()` resolves successfully;
- `SELECT COUNT(*) FROM usage_event` returns `0`;
- no detail event exists to be included in App Usage, export, or a future daily rollup.

Retrying the same cost-only payload does not repair the omission because the same early return is taken every time.

## Evidence

Relevant implementation:

- `SessionUsageSummary` declares every token field and `costAmount` optional: `pkgs/ag-ui-session/src/live-state.ts:114-132`.
- The zero-token early return happens before provided-cost handling: `apps/api/src/modules/cost/application/cost-usage-event.service.ts:126-162`.
- Unknown pricing intentionally uses `providedCostUsd` as the total: `apps/api/src/modules/cost/domain/cost-pricing.ts:476-498`.
- `usage_event` is the ledger consumed by the Cost query path: `docs/architecture.md:191-194` and `apps/api/src/modules/cost/application/cost-query-window.ts:101-152`.
- The product contract says unknown models may retain a reported USD amount: `docs/prd/cost-dashboard.md:24-26`.

I searched the current open and closed issue list, including #286, #305, #287/#317, #318, and #319. None covers cost-only usage being dropped before ledger persistence. #286 concerns rollup retention; #305 concerns duplicate public API mutation execution.

## Environment

- Mosoo commit: `e6401855d0b20a2b2bcf12d9a14659e443176d2b`
- Real cost usage service invocation with a deterministic local payload injection
- No remote Cloudflare account, Worker/D1 deployment, provider credential, provider request, Queue, or billable resource was used

This validates the local ledger persistence boundary. It does not claim that a remote production incident was observed.

## Contribution

I can submit a PR.

## Checklist

AsperforMias:
## Summary

- Persist a valid non-negative reported USD cost when all runtime token counters are absent.
- Preserve zero token buckets while retaining the reported amount for both known and unknown models.
- Record an explicit `runtime_reported_usd` pricing snapshot for known-model cost-only events.
- Keep the existing `(source, source_event_id)` upsert boundary so retries update one logical usage fact.
- Fixes #321.

## Why

`SessionUsageSummary` permits an upstream runtime or aggregator to report billed USD cost without a token breakdown. The previous zero-token early return discarded that valid cost before pricing and ledger persistence, while `session_model_call` could retain the same amount. This made App Usage, exports, detail queries, and later rollups undercount consumption.

This PR closes that ingestion gap without changing token-present pricing behavior. A separate follow-up PR on `fix/cost-ledger-recovery` is stacked on this branch and makes model-call plus ledger persistence atomic.

## Verification

- Commands:
  - `just test-package @mosoo/api` — 964 passed, 0 failed across 165 files.
  - `just tc-package @mosoo/api` — passed.
  - `just lint` — passed, including formatting checks and generated Worker type validation.
  - `git diff --check upstream/main...HEAD` — passed.
- Manual steps:
  - Invoked the real cost usage service against the local SQLite D1 fixture with `costAmount: 0.42`, `costCurrency: "USD"`, a declared usage contract, and no token counters.
  - Verified one `usage_event` row with zero token buckets and `420000` USD micros.
  - Replayed the same native call identity with an updated amount and verified the unique source boundary retained exactly one row.
  - Covered both unknown-model fallback and known-model reported-USD snapshot behavior.
- Not run:
  - No remote Cloudflare Worker/D1 deployment, provider request, Queue, production database, or billable resource was used. Verification covers the real local service/repository/D1 boundary, not a remote production incident.

## Impact

- User/API/contract changes:
  - App Usage and exports now include valid USD-only runtime consumption that was previously omitted.
  - No GraphQL, HTTP, or shared payload shape changes.
  - Cost-only usage now follows the same declared-`usageContract` requirement as token-bearing usage; current official Driver mappings already provide it.
  - N/A. No schema, migration, generated file, submodule, or lockfile change.
- Env or config changes:
  - N/A.
- Risk and rollback:
  - Risk is limited to cost-ledger semantics for valid USD-only events. Token-present pricing remains unchanged.
  - Existing last-write-wins upsert behavior is intentionally preserved because a runtime may update one logical usage fact.
  - Roll back this single commit to restore the former behavior; no data migration is required.

## Review

- Closest review areas:
  - Cost usage ingestion, usage-contract normalization, pricing fallback, and ledger idempotency.
- Known trade-offs:
  - Only finite, non-negative `USD` reports qualify as a reported-cost fallback.
  - This PR fixes cost-only ingestion but not the separate two-write atomicity boundary; the stacked follow-up PR depends on this one and addresses #322.

### DIFF
--- a/apps/api/src/modules/cost/application/cost-usage-event.service.ts
+++ b/apps/api/src/modules/cost/application/cost-usage-event.service.ts
@@ -131,12 +131,14 @@ export async function recordRuntimeUsageEvent(
   const rawOutputTokens = toTokenCount(input.usage.outputTokens);
   const rawCacheReadTokens = toTokenCount(input.usage.cachedReadTokens);
   const rawCacheCreationTokens = toTokenCount(input.usage.cachedWriteTokens);
+  const providedCostUsd = toProvidedUsdCost(input.usage);
 
   if (
     rawInputTokens === 0 &&
     rawOutputTokens === 0 &&
     rawCacheReadTokens === 0 &&
-    rawCacheCreationTokens === 0
+    rawCacheCreationTokens === 0 &&
+    providedCostUsd === null
   ) {
     return;
   }
@@ -158,7 +160,7 @@ export async function recordRuntimeUsageEvent(
     model,
     outputTokens: tokens.outputTokens,
     pricedAtMs: input.run.createdAtMs,
-    providedCostUsd: toProvidedUsdCost(input.usage),
+    providedCostUsd,
     provider,
   });
   const source = "runtime_driver";
--- a/apps/api/src/modules/cost/domain/cost-pricing.ts
+++ b/apps/api/src/modules/cost/domain/cost-pricing.ts
@@ -488,6 +488,29 @@ export function calculateUsageCost(input: {
     modelId: input.model,
     providerId: input.provider,
   });
+  const tokensUnavailable =
+    input.cacheCreationTokens === 0 &&
+    input.cacheReadTokens === 0 &&
+    input.inputTokens === 0 &&
+    input.outputTokens === 0;
+
+  if (tokensUnavailable && input.providedCostUsd !== null && input.providedCostUsd !== undefined) {
+    return {
+      priceSnapshotJson:
+        pricing === null
+          ? null
+          : JSON.stringify({
+              model: pricing.model,
+              provider: pricing.provider,
+              reportedCostUsd: input.providedCostUsd,
+              source: "runtime_reported_usd",
+              tokenCountersUnavailable: true,
+            }),
+      pricing,
+      pricingStatus: pricing === null ? "unknown" : "priced",
+      totalCostUsd: input.providedCostUsd,
+    };
+  }
 
   if (!pricing) {
     return {
--- a/apps/api/tests/cost-usage-event.test.ts
+++ b/apps/api/tests/cost-usage-event.test.ts
@@ -367,4 +367,114 @@ describe("cost usage event", () => {
       run_purpose: "channel",
     });
   });
+
+  test("persists and idempotently updates reported USD cost without token counters", async () => {
+    const database = createUsageEventDatabase();
+    const usage = {
+      costAmount: 0.42,
+      costCurrency: "USD",
+      source: "session_update",
+      usageContract: "openai_total_with_cached_breakdown",
+    } satisfies SessionUsageSummary;
+    const input = {
+      callKey: "cost-only-call",
+      driverInstanceId: DRIVER_INSTANCE_ID,
+      nativeCallId: "cost-only-native-call",
+      run: RUN_CONTEXT,
+      usage,
+    };
+
+    await recordRuntimeUsageEvent(database, input);
+    await recordRuntimeUsageEvent(database, {
+      ...input,
+      usage: {
+        ...usage,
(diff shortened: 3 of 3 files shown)


Yevanchen (approved):
Reviewed the diff against main.

- `toProvidedUsdCost` correctly gates the fallback to finite, non-negative, USD-only amounts; the early return now only skips events with neither tokens nor a valid reported cost.
- The new `tokensUnavailable` branch in `calculateUsageCost` only fires when all four token buckets are zero, so token-present pricing is unchanged. The `runtime_reported_usd` snapshot for known models keeps the audit trail explicit.
- The `(source, source_event_id)` upsert boundary is preserved, so retries still converge to one row.
- Tests cover unknown-model persistence + idempotent update and the known-model reported-USD snapshot.

CI (full `bun run check` incl. tests) is green. LGTM.

Issue #323, opened 2026-07-14
fix(db): audit and reconcile historical model-call ledger gaps


AsperforMias:
## Area

Database

## Summary

**Severity: P2 - conditional historical cost-ledger correctness.**

Before the atomic persistence fix in #322, the runtime path committed `session_model_call` and `usage_event` as two independent D1 writes. A failure of the second write could therefore leave a durable model-call fact without its corresponding usage-ledger fact. #322 prevents this partial state for future writes, but it is intentionally not a historical data migration and does not inspect or repair rows created before that fix.

If an affected deployment experienced the second-write failure and never received an equivalent runtime replay, App Usage, CSV exports, seven-day detail queries, and daily rollups can continue to understate consumption after the code upgrade. There is currently no operator-facing audit that distinguishes a valid model call with a missing ledger row from a fully persisted call.

This report does **not** claim that a remote Mosoo or Cloudflare production database contains such rows. The state is deterministically constructible from the former write ordering, but confirming its presence and magnitude in a particular deployment requires a read-only audit of that deployment.

Reconciliation must be conservative. Raw `usage_event` rows are normally rolled up and deleted after seven days. Consequently, a historical `session_model_call` without a raw `usage_event` is not automatically an omission: it may already be represented in `usage_daily_rollup`. Blindly backfilling every anti-join result could count the same usage twice. Missing immutable historical dimensions can also make an old row unsafe to reconstruct.

The financial impact is serious even though App Usage is currently an operational estimate rather than an invoice. Undercounting misleads spend analysis; an unsafe repair can overcount consumption. If this ledger later drives quotas, credits, chargeback, or billing, both failure modes become P1.

## Steps to reproduce

1. Check out a revision before #322, such as `e4be8e7753502c46f8c58f483a7b47f811617cd9`, and create the local SQLite D1 fixture with a valid App, Agent, Session, Session Run, `session_model_call`, and `usage_event` schema.
2. Invoke the real `upsertSessionModelCallUsage()` path with a stable `driverInstanceId`, `callId`, non-zero token usage, and a declared usage contract.
3. Inject a failure only for the first `INSERT INTO usage_event ... ON CONFLICT ...` statement, after allowing the earlier `session_model_call` upsert to commit.
4. Confirm the former behavior:

   ```sql
   SELECT COUNT(*) FROM session_model_call; -- 1
   SELECT COUNT(*) FROM usage_event;        -- 0
   ```

[...]

PR #325, opened 2026-07-14
fix(cost): atomically persist usage ledger

### LINKED ISSUE #322: fix(cost): make model-call and usage-ledger persistence recoverable

AsperforMias:
## Summary

- Commit `session_model_call` and its corresponding `usage_event` in one D1 batch.
- Extract the usage-ledger upsert builder so the existing standalone service and the model-call transaction share identical pricing and source-identity semantics.
- Preserve both unique retry boundaries: `(session_run_id, call_key)` for model calls and `(source, source_event_id)` for usage events.
- Add a D1 boundary-injection regression proving failure rollback and exactly-once retry convergence.
- Fixes #322.

**Dependency: #324 must merge first.** This branch is intentionally stacked on `fix/cost-reported-usage`; until #324 lands, GitHub will show that prerequisite commit in this PR's comparison.

## Why

The previous runtime persistence sequence committed `session_model_call` before writing `usage_event`. A transient failure on the second write left contradictory durable state: the model call and its usage could exist in one table while App Usage, exports, and rollups omitted it.

A single D1 batch removes that partial-commit state. If either statement fails, both roll back; replaying the same runtime event then converges through the existing unique keys without duplicate cost.

This is a forward consistency fix. It does not guess at or mutate possible historical partial rows created before deployment. The separate, retention-aware audit and reconciliation boundary is tracked in #323.

## Verification

- Commands:
  - `just test-package @mosoo/api` — 965 passed, 0 failed across 165 files.
  - `just tc-package @mosoo/api` — passed.
  - `just lint` — passed, including formatting checks and generated Worker type validation.
  - `just commit-check` — passed for both stacked Conventional Commits.
  - `git diff --check fix/cost-reported-usage...HEAD` — passed.
- Manual steps:
  - Invoked the real `upsertSessionModelCallUsage()` repository/service path against the local SQLite D1 fixture.
  - Injected a failure only on the first `usage_event` insert inside the D1 batch.
  - Verified the rejected attempt left zero `session_model_call` rows and zero `usage_event` rows.
  - Retried the identical input twice and verified exactly one model-call row and one usage-ledger row.
  - Verified runtime event receipts remain downstream of this transaction, so a later receipt failure replays through the same idempotent upserts.
- Not run:
  - No remote Cloudflare Worker/D1 deployment, provider request, Queue, production database, or billable resource was used. The local adapter executes the real repository statements in a transaction and the injected test validates the CF D1 batch boundary contract locally; this is not a claim of observing a remote production incident.

## Impact

- User/API/contract changes:
  - No public GraphQL, HTTP, or shared payload shape changes.
  - Successful model-call persistence produces the same rows and pricing values as before.
  - On a ledger-write failure, the model-call row now rolls back instead of remaining as a contradictory diagnostic record. This failure-state change is necessary to preserve cost consistency.
  - N/A. No schema, migration, generated file, submodule, or lockfile change.
- Env or config changes:
  - N/A.
- Risk and rollback:
  - The primary risk is coupling two writes into one transaction, increasing the scope of a statement failure; the intended result is fail-closed with neither fact committed.
  - Session/Run foreign-key failure also aborts the entire batch, so an orphan usage row cannot be created through this path.
  - Roll back the top commit after #324 to restore the former sequential writes; no migration is required.
  - Historical pre-fix gaps are not backfilled by this PR and are explicitly tracked in #323.

## Review

- Closest review areas:
  - Runtime event persistence ordering, D1 batch semantics, model-call identity, cost-ledger source identity, and retry behavior.
- Known trade-offs:
  - The Run context read still occurs before the write batch; Session/Run foreign keys make a concurrent deletion fail and roll back both writes.
  - Existing last-write-wins updates for repeated usage payloads remain unchanged.
  - This PR depends on #324 and should be reviewed as the single top commit once that prerequisite merges.

### DIFF
--- a/apps/api/src/modules/cost/application/cost-usage-event.service.ts
+++ b/apps/api/src/modules/cost/application/cost-usage-event.service.ts
@@ -16,6 +16,7 @@ import type {
 import { sql } from "drizzle-orm";
 
 import { getAppDatabase } from "../../../platform/db/drizzle";
+import type { AppDatabase } from "../../../platform/db/drizzle";
 import { isTruthy } from "../../../shared/truthiness";
 import { calculateUsageCost } from "../domain/cost-pricing";
 import { normalizeUsageTokens } from "../domain/usage-contract";
@@ -123,10 +124,10 @@ function toUsdMicros(value: number): number {
   return Math.round(value * 1_000_000);
 }
 
-export async function recordRuntimeUsageEvent(
-  database: D1Database,
+export function createRuntimeUsageEventUpsert(
+  database: AppDatabase,
   input: RecordRuntimeUsageEventInput,
-): Promise<void> {
+) {
   const rawInputTokens = toTokenCount(input.usage.inputTokens);
   const rawOutputTokens = toTokenCount(input.usage.outputTokens);
   const rawCacheReadTokens = toTokenCount(input.usage.cachedReadTokens);
@@ -140,7 +141,7 @@ export async function recordRuntimeUsageEvent(
     rawCacheCreationTokens === 0 &&
     providedCostUsd === null
   ) {
-    return;
+    return null;
   }
 
   const provider = input.run.provider;
@@ -168,7 +169,7 @@ export async function recordRuntimeUsageEvent(
     ? `${input.driverInstanceId}:${input.nativeCallId}`
     : `${input.driverInstanceId}:${input.run.sessionRunId}:${input.callKey}`;
 
-  await getAppDatabase(database)
+  return database
     .insert(usageEventsTable)
     .values({
       actorUserId: input.run.actorUserId,
@@ -211,6 +212,18 @@ export async function recordRuntimeUsageEvent(
         usageContract: sql`excluded.usage_contract`,
       },
       target: [usageEventsTable.source, usageEventsTable.sourceEventId],
-    })
-    .run();
+    });
+}
+
+export async function recordRuntimeUsageEvent(
+  database: D1Database,
+  input: RecordRuntimeUsageEventInput,
+): Promise<void> {
+  const query = createRuntimeUsageEventUpsert(getAppDatabase(database), input);
+
+  if (query === null) {
+    return;
+  }
+
+  await query.run();
 }
--- a/apps/api/src/modules/sessions/infrastructure/session-model-call.repository.ts
+++ b/apps/api/src/modules/sessions/infrastructure/session-model-call.repository.ts
@@ -20,10 +20,10 @@ import type {
 } from "@mosoo/id";
 import { and, eq, sql } from "drizzle-orm";
 
-import { getAppDatabase } from "../../../platform/db/drizzle";
+import { getAppDatabase, runAppDatabaseBatch } from "../../../platform/db/drizzle";
 import { isTruthy } from "../../../shared/truthiness";
 import { currentTimestampMs } from "../../../time";
-import { recordRuntimeUsageEvent } from "../../cost/application/cost-usage-event.service";
+import { createRuntimeUsageEventUpsert } from "../../cost/application/cost-usage-event.service";
 import type { SessionUsageSummary } from "./session-live-state.types";
 interface SessionModelCallRunRow {
   agent_id: AgentId;
@@ -149,6 +149,7 @@ export async function upsertSessionModelCallUsage(
   if (!input.usage) {
     return;
   }
+  const usage = input.usage;
 
   const run = await getSessionModelCallRunRow(database, input.sessionRunId);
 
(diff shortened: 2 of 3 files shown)


Yevanchen (approved):
Reviewed the incremental diff on top of #324.

- The refactor to `createRuntimeUsageEventUpsert` is behavior-preserving: same values, same `(source, source_event_id)` conflict clause; the standalone `recordRuntimeUsageEvent` wrapper keeps existing callers intact.
- `runAppDatabaseBatch` (existing helper over drizzle's D1 `batch()`) gives the atomic boundary; both unique retry keys — `(session_run_id, call_key)` and `(source, source_event_id)` — are preserved so replays converge.
- One intentional behavior change worth noting: a missing usage contract now throws while *building* the batch, so the model-call row also rolls back (previously it would have committed before `recordRuntimeUsageEvent` threw). That is consistent with the atomicity goal of this PR — fail-closed with neither fact committed.
- The injected-failure regression test proves rollback (0 rows in both tables) and exactly-once convergence after two retries.

CI (full `bun run check` incl. tests) is green. LGTM.

PR #326, opened 2026-07-14
fix(db): reconcile historical model call ledger gaps in consistency

### LINKED ISSUE #323: fix(db): audit and reconcile historical model-call ledger gaps

AsperforMias:
## Summary

- Add an opt-in, bounded audit and reconciliation workflow for historical `session_model_call` rows that lack their canonical `usage_event` ledger fact.
- Default the service to read-only `audit`; allow explicit `repair` only for recent rows whose source identity, usage values, published revision, and Run/Session/App context can be proven.
- Preserve the existing `(source, source_event_id)` uniqueness boundary and use `ON CONFLICT DO NOTHING` so reconciliation retries converge without replacing a concurrent runtime write.
- Run pages through the durable API command ledger and Queue cursor chain, with structured counts and a retention cutoff recalculated at actual processing time.
- Keep records outside the seven-day raw-detail window, ambiguous historical-driver identities, invalid values, and incomplete context as `indeterminate` instead of guessing or rewriting daily aggregates.
- Make the existing daily rollup raw statement executable inside the current Drizzle D1 batch implementation without changing its cutoff or aggregate-then-delete semantics.
- Fixes #323.

**Dependencies: #324 and #325 must merge first, in that order.** This branch is intentionally stacked on `fix/cost-ledger-recovery` (#325), which is itself stacked on `fix/cost-reported-usage` (#324). Until both prerequisites land, GitHub will show their commits in this PR comparison. The intended merge order is **#324 -> #325 -> this PR**.

## Why

Before #325, model-call persistence and usage-ledger persistence were separate D1 writes. If the first write committed and the second failed, the durable state could contain a valid `session_model_call` without the corresponding cost-ledger row. #325 prevents that partial state for future writes, but it deliberately does not inspect or repair historical rows.

A blind anti-join backfill is unsafe. Raw `usage_event` detail is rolled into `usage_daily_rollup` and deleted after seven days, so an old model call without raw detail may already be represented by an aggregate. A model call can also have an event under a historical driver identity, or lack immutable context needed to reproduce the original dimensions. Reconstructing these cases could double count consumption.

This PR therefore limits mutation to recently created calls that are still inside the same raw-detail retention boundary and can be reconstructed through the canonical runtime ingestion rules. Everything else remains observable but fail-closed. The workflow is disabled by default and operators are expected to run `audit` before enabling `repair`.

## Verification

- Commands:
  - `just check` - passed: formatting, documentation links, lint, all workspace typechecks and tests, and GraphQL codegen freshness. The API suite reported 980 passed, 0 failed, and 3380 assertions across 166 files.
  - `just commit-check` - passed for all 7 stacked commits from `origin/main` through this branch.
  - `git diff --check fix/cost-ledger-recovery...HEAD` - passed.
  - Focused reconciliation suite - 15 passed, 0 failed, and 46 assertions.
- Manual steps:
  - Seeded the real local SQLite D1 repository shape with the historical `session_model_call = 1` / `usage_event = 0` state from the issue reproduction.
  - Verified default audit classifies the gap without mutation, explicit repair creates exactly one canonical event, and a second repair observes the existing event without duplication.
  - Verified native-call and Run/call-key fallback source identities.
  - Injected a failure on the second statement of a real D1 repair batch; verified the first insert rolled back and retry repaired both rows exactly once.
  - Injected a concurrent runtime ledger write between the audit query and repair batch; verified reconciliation did not overwrite its amount or dimensions.
  - Ran the real `runUsageDailyRollup()` path, verified old raw detail moved into `usage_daily_rollup` and was deleted, then verified reconciliation neither resurrected the raw event nor changed the aggregate.
  - Exercised the durable API command ledger and local Queue boundary, including a command delayed beyond retention and cursor continuation past a corrupt model-call ID.
  - Verified historical driver ambiguity, pre-retention call creation, invalid stored usage values, malformed channel metadata, missing revision, and missing context remain non-mutating `indeterminate` states.
- Not run:
  - No remote Cloudflare Worker/D1 deployment, production database read, provider request, production Queue, or billable Cloudflare resource was used.
  - Verification uses the real local service/repository/SQLite D1 path plus injected D1 batch, Queue, concurrency, and Cloudflare boundary failures. It proves local behavior and retry semantics, not that a particular remote production database contains affected rows.

## Impact

- User/API/contract changes:
  - No public GraphQL, HTTP, or shared client payload changes.
  - App Usage, exports, and later rollups can recover only locally provable recent historical omissions.
  - Audit and repair results are emitted as structured logs with scanned, present, repairable, repaired, skipped, indeterminate, and failed information.
  - No generated GraphQL files, D1 migration files, frozen baseline changes, dependency changes, lockfile changes, or submodule changes.
  - `ApiCommandKind` gains a TypeScript command value only; the underlying D1 column is already text and requires no DDL migration.
- Env or config changes:
  - Adds optional `MOSOO_COST_LEDGER_RECONCILIATION_MODE` with exact values `audit` or `repair`.
  - Unset or blank keeps the workflow disabled. When enabled, a new run starts at 01:00 UTC and pages continue through durable Queue commands.
  - Operators should enable `audit`, inspect page logs, switch to `repair` only after review, and unset the variable after the intended run.
- Risk and rollback:
  - Cost consistency is treated fail-closed: uncertain rows are never written, records older than raw retention are never backfilled, and reconciliation never updates an existing daily aggregate.
  - Repair inserts use one D1 atomic batch per bounded page and do not overwrite a concurrent live event.
  - The daily rollup compatibility adjustment preserves its existing seven-day cutoff and atomic aggregate-then-delete behavior; it only avoids the current Drizzle D1 limitation for parameterized `db.run(sql)` batch entries.
  - Code rollback is straightforward by reverting these five top commits. Any ledger rows already inserted by an explicitly enabled repair are durable business records and are intentionally not deleted by code rollback; reversing those would require a targeted, reviewed data correction.

## Review

- Closest review areas:
  - Historical source-identity reconstruction and ambiguity handling.
  - Seven-day detail retention versus daily-rollup boundaries.
  - D1 batch atomicity, uniqueness conflicts, Queue retry/continuation, and actual-processing-time cutoffs.
  - Classification of published revision, Session/Run/App context, and malformed historical data.
- Known trade-offs:
  - History before raw retention is reported as an indeterminate-presence boolean rather than counted or repaired, avoiding an unsafe full-history financial rewrite.
  - The first opt-in page performs a read-only old-history existence probe whose read cost is data-volume dependent; repair writes and cursor pages remain bounded to at most 100 candidates.
  - This workflow reconciles Mosoo's internal operational usage ledger, not provider invoices, taxes, credits, discounts, or infrastructure charges.
  - This PR must not merge before #324 and #325.

### DIFF
--- a/apps/api/src/modules/api-command/application/api-command-enqueue.ts
+++ b/apps/api/src/modules/api-command/application/api-command-enqueue.ts
@@ -5,6 +5,7 @@ import { enqueueApiCommand } from "./api-command-ledger";
 import type {
   AppDeploymentRunDispatchCommandPayload,
   ChannelWorkTriggerCommandPayload,
+  CostLedgerReconciliationCommandPayload,
   ScheduledMaintenanceCommandPayload,
   SessionRunDispatchCommandPayload,
 } from "./api-command-payload";
@@ -37,6 +38,22 @@ export async function enqueueChannelWorkTriggerCommand(
   await enqueueApiCommand(bindings, {
     dedupeKey: createChannelWorkTriggerDedupeKey(payload),
     kind: "channel_work_trigger",
+    payload,
+  });
+}
+
+export async function enqueueCostLedgerReconciliationCommand(
+  bindings: Pick<ApiBindings, "API_COMMAND_QUEUE" | "DB">,
+  payload: CostLedgerReconciliationCommandPayload,
+): Promise<void> {
+  await enqueueApiCommand(bindings, {
+    dedupeKey: [
+      "cost_ledger_reconciliation",
+      payload.scheduledTime,
+      payload.mode,
+      payload.cursor ?? "start",
+    ].join(":"),
+    kind: "cost_ledger_reconciliation",
     payload,
   });
 }
--- a/apps/api/src/modules/api-command/application/api-command-payload.ts
+++ b/apps/api/src/modules/api-command/application/api-command-payload.ts
@@ -15,10 +15,15 @@ import type { DiscordWorkTrigger } from "../../channels/discord/discord-events";
 import type { LarkWorkTrigger } from "../../channels/lark/lark-events";
 import type { SlackWorkTrigger } from "../../channels/slack/slack-events";
 import type { TelegramWorkTrigger } from "../../channels/telegram/telegram-events";
+import type {
+  CostLedgerReconciliationCursor,
+  CostLedgerReconciliationMode,
+} from "../../cost/application/cost-ledger-reconciliation.service";
 
 type ApiCommandPayload =
   | AppDeploymentRunDispatchCommandPayload
   | ChannelWorkTriggerCommandPayload
+  | CostLedgerReconciliationCommandPayload
   | ScheduledMaintenanceCommandPayload
   | SessionRunDispatchCommandPayload;
 
@@ -53,6 +58,12 @@ export type ChannelWorkTriggerCommandPayload =
     };
 
 export interface ScheduledMaintenanceCommandPayload {
+  scheduledTime: number;
+}
+
+export interface CostLedgerReconciliationCommandPayload {
+  cursor: CostLedgerReconciliationCursor | null;
+  mode: CostLedgerReconciliationMode;
   scheduledTime: number;
 }
 
@@ -436,6 +447,35 @@ function parseScheduledMaintenancePayload(value: unknown): ScheduledMaintenanceC
   };
 }
 
+function parseCostLedgerReconciliationPayload(
+  value: unknown,
+): CostLedgerReconciliationCommandPayload {
+  const label = "cost_ledger_reconciliation payload";
+  const record = requireRecord(value, label);
+  const cursor = readOptionalString(record, "cursor", label);
+  const mode = readNonEmptyString(record, "mode", label);
+
+  if (mode !== "audit" && mode !== "repair") {
+    throw new ApiCommandPayloadError(`${label}.mode must be 'audit' or 'repair'.`);
+  }
+
+  if (cursor !== null && cursor.length === 0) {
+    throw new ApiCommandPayloadError(`${label}.cursor must not be empty.`);
+  }
+
+  const scheduledTime = readInteger(record, "scheduledTime", label);
+
+  if (scheduledTime < 0 || !Number.isFinite(new Date(scheduledTime).getTime())) {
+    throw new ApiCommandPayloadError(`${label}.scheduledTime must be a valid timestamp.`);
+  }
+
+  return {
+    cursor,
+    mode,
+    scheduledTime,
+  };
+}
+
 function parseAppDeploymentRunDispatchPayload(
   value: unknown,
 ): AppDeploymentRunDispatchCommandPayload {
@@ -469,6 +509,9 @@ export function parseApiCommandPayload(
     }
     case "channel_work_trigger": {
       return parseChannelWorkTriggerPayload(parsed);
+    }
+    case "cost_ledger_reconciliation": {
+      return parseCostLedgerReconciliationPayload(parsed);
     }
     case "scheduled_maintenance": {
       return parseScheduledMaintenancePayload(parsed);
--- a/apps/api/src/modules/api-command/application/api-command-processor.ts
+++ b/apps/api/src/modules/api-command/application/api-command-processor.ts
@@ -29,10 +29,17 @@ import { processLarkWorkTrigger } from "../../channels/lark/lark-first-party-ada
 import { processSlackWorkTrigger } from "../../channels/slack/slack-first-party-adapter";
 import { parseTelegramCredentials } from "../../channels/telegram/telegram-credentials";
 import { processTelegramWorkTrigger } from "../../channels/telegram/telegram-first-party-adapter";
+import {
+  parseCostLedgerReconciliationActivationMode,
+  reconcileCostLedgerPage,
+} from "../../cost/application/cost-ledger-reconciliation.service";
 import { runUsageDailyRollup } from "../../cost/application/cost-rollup.service";
 import { dispatchQueuedSessionRun } from "../../runtime/application/session-runs/dispatch-queued-run.service";
 import { runSandboxMaintenance } from "../../runtime/infrastructure/runtime-subject-lifecycle/runtime-subject-maintenance.service";
-import { APP_DEPLOYMENT_RUN_DISPATCH_DEDUPE_PREFIX } from "./api-command-enqueue";
+import {
+  APP_DEPLOYMENT_RUN_DISPATCH_DEDUPE_PREFIX,
+  enqueueCostLedgerReconciliationCommand,
+} from "./api-command-enqueue";
 import {
   API_COMMAND_LEASE_RENEWAL_INTERVAL_MS,
   claimApiCommand,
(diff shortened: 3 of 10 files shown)


Yevanchen (approved):
Reviewed the incremental diff on top of #325 in depth (service, shared ingestion path, api-command ledger/processor, rollup compatibility change, and the test suite).

Verified the core safety claims:

- Default is read-only `audit`; blank/unset `MOSOO_COST_LEDGER_RECONCILIATION_MODE` keeps the workflow disabled entirely, and audit mode performs zero writes.
- Repair is fail-closed: candidates require in-retention `created_at`, provable source identity, valid stored usage values, an existing published revision row, and full Run/Session/App/Org context — everything else stays `indeterminate`/skipped with no mutation. Rollup double-counting is excluded (a repairable candidate's hypothetical live event could not yet have been rolled up; the `predates_safe_repair_window` check kills the started-long-ago/completed-recently case).
- Repair reuses the exact live-ingestion insert builder with `ON CONFLICT DO NOTHING` on `(source, source_event_id)`, so retries converge and a concurrent runtime write is never overwritten (covered by tests).
- Pages are bounded (≤100), run as atomic D1 batches, and chain through durable api-commands with dedupe keys; the interrupted-batch and 8-day-delayed-command tests cover the retry semantics.
- The `cost-rollup.service.ts` change is behavior-preserving: same cutoff helper, `sql.raw` only interpolates a `Number.isSafeInteger`-guarded integer, so no injection surface.
- The `createRuntimeUsageEventInsert` refactor leaves live ingestion behavior unchanged.

Non-blocking findings, fine as follow-ups:

1. **api-command-processor.ts** — an invalid (non-blank, non-`audit`/`repair`) env value makes `parseCostLedgerReconciliationActivationMode` throw inside the scheduled-maintenance command after the always-on maintenance promises have started, so one operator typo repeatedly fails the whole 01:00 maintenance tick until dead-lettered. Catching the parse error and skipping only the reconciliation enqueue would fail closed without collateral damage.
2. **cost-ledger-reconciliation.service.ts** — in repair mode a candidate that loses the ON CONFLICT race is counted in both `repairable` and `present`, so the page-completed counters can sum to more than `scanned`. Worth a separate `conflicted` counter since operators are told to read these logs before enabling repair.
3. The first-page pre-retention history probe and the retention-window scan use non-indexable predicates (`COALESCE(completed_at, created_at)`), so on a large `session_model_call` table the nightly run degrades to a near-full scan in one D1 statement. Bounded per run, but worth an index or documented size expectation.

CI (full `bun run check` incl. tests) is green; the focused reconciliation suite passes locally. LGTM.

Issue #329, opened 2026-07-15
fix(api): atomically admit Session Runs and dispatch commands


AsperforMias:
## Area

API

## Summary

**Severity: P2 - partial Session Run admission can lose a request and block the Session.**

`queueSessionRun()` commits the active `session_run`, user `session_message`, queued `session_event` rows, and durable `api_command` in separate D1 operations. The durable Queue outbox introduced by #299 begins only after the `api_command` row exists.

A transient D1 failure at any later write point therefore leaves a different half-completed state. In every reproduced case the Run remains active with status `queued`; an immediate retry is rejected because that Run occupies the Session, but no complete durable admission exists to execute or replay the original request.

The stale-run sweep can eventually mark the orphan Run failed, but it cannot reconstruct the missing message, events, or dispatch command. Production impact includes lost user input, Sessions blocked until reconciliation, contradictory timeline/state views, and an accepted-looking queued Run that can never dispatch.

## Steps to reproduce

1. Check out current `main`, which already includes #299.
2. Create the local SQLite D1 public API fixture with a valid owner, App, Agent, idle Session, provider configuration, and API-command Queue stub.
3. Call the real `queueSessionRun()` service with one prompt, no attachments, and a stable `clientRequestId`.
4. Wrap D1 so exactly one selected insert throws, while all preceding statements execute normally.
5. Inject the failure at `INSERT INTO session_message`, then query `session_run`, `session_message`, `session_event`, and `api_command`.
6. Repeat from a fresh fixture with failure at `INSERT INTO session_event`.
7. Repeat from a fresh fixture with failure at `INSERT INTO api_command`.
8. After each failure, immediately retry the same `queueSessionRun()` request.

Observed durable states:

| Injected failure | session_run | session_message | session_event | api_command |
| --- | ---: | ---: | ---: | ---: |
| `session_message` | 1 queued | 0 | 0 | 0 |
| `session_event` | 1 queued | 1 | 0 | 0 |
| `api_command` | 1 queued | 1 | >0 | 0 |

Every immediate retry fails with `This conversation already has an active run`.

## Expected behavior

Admission of one Session Run must expose one atomic durable outcome:

- either the Run, user message, canonical queued events, and durable dispatch obligation are all committed; or
- none of them are committed and the caller can safely retry.

[...]

PR #338, opened 2026-07-15
fix(api): make bound agent calls idempotent

### LINKED ISSUE #327: fix(api): make bound Agent calls idempotent across ambiguous retries
## Area

API

## Summary

**Severity: P1 - duplicate owner-billed execution after an ambiguous HTTP result.**

`POST /api/v1/bound/:token` does not participate in the public API idempotency boundary and starts every accepted request with `clientRequestId: null`. If the endpoint has already created a Session, queued a Run, and persisted its durable dispatch command but the caller receives a timeout or loses the response, retrying the identical token and body creates a second Session and Run.

The shared rate limiter bounds request volume but cannot identify retries of one logical request. In production, both accepted Runs are independently dispatchable and can execute the same Agent work against the App owner's provider credential, causing duplicate side effects and real duplicate model spend.

This report does **not** claim that a remote Mosoo deployment has observed duplicate charges. The duplicate durable admission was reproduced locally through the real bound-call service, D1 repositories, and API-command enqueue path without invoking a model provider.

## Steps to reproduce

1. Use the local public HTTP/D1 fixture with a published Agent, valid App, successful Deployment/DeploymentRun, matching public binding, and a valid bound capability token.
2. Use the real `createBoundAgentThreadAndWait()` service called by `POST /api/v1/bound/:token`, with the normal D1 and API-command Queue stub. Set `executionContext: null` so no provider execution is needed.
3. Submit `{ "message": "same logical request" }` with the capability token.
4. Advance the injected wait clock until the blocking call returns `deployment_agent_call_timeout` after Session/Run admission has succeeded.
5. Discard that result to model an ambiguous client-side timeout, then submit the exact same token and body again.
6. Query the durable rows:

   ```sql
   SELECT COUNT(*) FROM session;
   SELECT COUNT(*) FROM session_run;
   SELECT COUNT(*) FROM api_command;
   ```

7. Observe `2 / 2 / 2`, with two distinct Run IDs and both admissions carrying `clientRequestId = null`.

## Expected behavior

One logical bound request must have a durable idempotency identity across HTTP retries.

After the first request has admitted a Run, an identical retry with the same caller-provided idempotency key should either return/recover the original Run result or report that the original request is still in progress. It must not create a second Session, Run, dispatch command, or billable Agent execution.

The idempotency subject must be scoped to the verified bound capability/deployment identity rather than a PAT token, and mismatched reuse of a key must fail closed.

## Actual behavior

Both identical calls independently passed capability validation and rate limiting, created a new Session, queued a new Run, and inserted a new `session_run_dispatch` command before timing out.

The local assertions observed:

- first call: timeout after one durable Session/Run/command;
- identical retry: a second timeout after another durable Session/Run/command;
- final counts: `session = 2`, `session_run = 2`, `api_command = 2`.

## Evidence

- The bound route calls its service directly and does not use `runPublicApiIdempotentJson()`: `apps/api/src/adapters/http/routes/public-api-route.ts:106-123`.
- Bound Run admission passes `clientRequestId: null`: `apps/api/src/modules/public-api/app-agent-bound-ask.service.ts:87-128`.
- The rate-limit comment explicitly recognizes that the capability can launch owner-billed Runs, but rate limiting is not deduplication: `apps/api/src/modules/public-api/app-agent-bound-ask.service.ts:159-175`.
- Focused reproduction command:

  ```text
  just test-file apps/api/tests/tmp-bound-agent-idempotency-repro.test.ts
  1 pass, 0 fail, 5 assertions
  ```

The temporary test used the real service, SQLite-backed D1 fixture, capability verification, Session/Run repositories, and API-command Queue stub. A controlled clock compressed the normal 25-second wait; provider discovery was locally stubbed and no Run was remotely dispatched.

I searched current open and closed issues and PRs. #305/#312 cover completion-record failure for PAT-backed public Thread/Event routes that already opt into the shared idempotency wrapper. #287/#317 cover bound capability revocation, #318 covers external Cloudflare drift, and #319/#320 cover accepted-Run provenance. None provides retry idempotency for the bound endpoint.

## Environment

- Verified branch: `fix/bound-capability-revocation` (PR #317)
- Verified commit: `195f65e08290f8f3eb926b52e8a9f1c0ef2d4fed`
- Bun `1.4.0-canary.1`
- Local SQLite D1 fixture and API-command Queue stub
- No remote Cloudflare account, Worker, D1, Queue, capability URL, provider credential, model request, or billable resource

The implementation should follow #317 because it touches the same bound authorization/service path, but the correctness problem is independent of capability revocation.

## Contribution

I can submit a PR.

## Checklist

AsperforMias:
## Summary

- Add standard `Idempotency-Key` handling to `POST /api/v1/bound/:token`.
- Reserve a stable Session identity in D1 before Run admission, bind the original Run durably, and recover that exact Session/Run after timeout, lost response, or an interrupted binding write.
- Scope keys to the verified bound capability/deployment identity and reject same-key/different-body reuse with `409`.
- Preserve the existing behavior for callers that do not send an idempotency key.
- Add an additive D1 migration plus HTTP/service-level concurrency and fault-injection coverage.

Fixes #327.

## Why

- Before this change, an accepted bound call could create and queue a Run, return an ambiguous timeout to the caller, and create a second owner-billed Session/Run when the caller retried the same logical request.
- Rate limiting limits volume but cannot distinguish a retry from a new logical request. The durable reservation gives the request a stable identity before any billable Run is admitted.
- The original reproduction observed `2 Session / 2 Run / 2 api_command` rows after two identical timeout attempts. The same timeout trigger now leaves exactly `1 / 1 / 1` and one Queue delivery.

## Verification

- Commands:
  - `just test-file apps/api/tests/bound-agent-idempotency.e2e.test.ts` - 12 passed, 0 failed, 55 assertions.
  - `just tc-package @mosoo/db`
  - `just db-reset-local` - applied migrations `0000`, `0001`, and the new additive `0002` to local Wrangler D1 only.
  - `just check` - exit 0; formatting, docs, lint, all typechecks, all package tests, API 1013 passed/0 failed, and GraphQL codegen check passed.
- Manual steps:
  - Replayed the issue repro by forcing the first and second service calls to reach `deployment_agent_call_timeout` after durable admission, then queried D1 and confirmed one reservation, Session, Run, and dispatch command.
  - Sent concurrent identical HTTP requests and confirmed both resolve from one Session/Run and only one Queue message is sent.
  - Discarded a successful HTTP result and retried the same key/body to model an ambiguous response; the response and original Run were recovered.
  - Injected failures at Session creation, Run admission, and reservation-to-Run binding, then confirmed retries recover the reserved identity without duplicate admission.
  - Verified same-key/different-body rejection, capability subject isolation, revocation recheck, and legacy no-key behavior.
- Not run:
  - No remote Cloudflare account, deployed Worker, remote D1, Queue, Durable Object, provider credential, model request, or billable resource was used.
  - **Source-level invariants, the issue repro, and the local injected HTTP/service/D1/outbox chain demonstrate that the business-consistency defect is fixed. The remaining unverified boundary is remote Cloudflare platform integration behavior; it is not an absence of validation of the issue's business-consistency fix.**

## Impact

- User/API/contract changes:
  - Bound Agent callers may provide `Idempotency-Key`. Identical retries under the same verified capability/deployment identity recover the original work; conflicting reuse returns `409`.
  - Requests without the header retain the previous non-idempotent behavior for backward compatibility.
  - Adds `bound_agent_call_idempotency_key`, its unique subject/key index, and an updated-at index through additive migration `0002_bound-agent-call-idempotency.sql`.
  - Includes the generated Drizzle snapshot/journal and local HTTP fixture schema update.
  - No destructive SQL, GraphQL output, or lockfile changes.
- Env or config changes:
  - None.
- Risk and rollback:
  - Main risks are reservation recovery and subject scoping; concurrency, timeout, ambiguous-response, persistence-failure, body-conflict, and revocation cases are covered directly.
  - Reverting the application commit restores the previous route behavior. The additive table can remain unused safely; production rollback does not require destructive D1 SQL.

## Review

- Closest review areas:
  - Public bound-call authentication/rate-limit/idempotency ordering.
  - Stable Session/Run recovery after interrupted writes.
  - D1 schema, migration, and capability/deployment subject scope.
- Known trade-offs:
  - Idempotency is opt-in through the caller-provided header; silently deduplicating unkeyed requests would conflate legitimate repeated prompts.
  - Reservations are retained durably so an old accepted request cannot later be executed twice. Retention/GC is intentionally not introduced in this correctness fix.
  - This PR can merge independently of #329; #329 hardens atomic Run admission beneath this endpoint and addresses a separate partial-write failure mode.


AsperforMias:
那个七千多lines diff的是drizzle为每次迁移生成的完整db schema snapshot

PR #347, opened 2026-07-16
fix(threads): show run failure details

RockChinQ:
## Summary

- show persisted last-run failure details inline on Thread detail
- make the failed status and inline notice open the process dialog
- add status-specific fallbacks and regression tests

## Why

A run can fail before the Agent produces a reply. The backend already persists `lastRun.error` and the `run.failed` event, but Thread detail rendered only a `Failed` badge. The process dialog entry also existed only on Agent reply cards, so pre-ready failures had no visible explanation or way to inspect the process.

The reproduced run failed during Driver provisioning with `runtime.provision_failed` and `Driver process exited before ready with exit code 1.` This change surfaces that persisted user-safe error without exposing raw Driver logs.

## Verification

- `bun run --filter @mosoo/web lint`
- `bun run --filter @mosoo/web tc`
- `bun run --filter @mosoo/web test` (158 passed)
- `vp fmt --check` on the three changed files
- `git diff --check`
- `just commit-check`
- Playwright: verified the original failed Thread shows the error inline and opens the process dialog
- Playwright: verified desktop and 390px mobile layouts

## Impact

- User/API/contract changes: failed Threads now show persisted run details; no API contract changes
- Generated files / GraphQL / DB / lockfile: none
- Env or config changes: none
- Risk and rollback: Web-only presentation change; revert this commit to restore the previous badge-only behavior

## Review

- Closest review: failure-state mapping and Thread detail process interaction
- Known trade-off: only persisted user-safe `lastRun.error` is displayed; raw Driver stderr remains in operator logs

### DIFF
--- a/apps/web/src/routes/threads/detail/run-failure-notice.tsx
+++ b/apps/web/src/routes/threads/detail/run-failure-notice.tsx
@@ -1,0 +1,92 @@
+import type { SessionRunStatus, SessionRunSummary } from "@mosoo/contracts/session-run";
+import { ChevronRight, CircleX } from "lucide-react";
+import type { ReactElement } from "react";
+
+import { Button } from "@/shared/ui/button";
+
+interface ThreadRunFailureDetails {
+  code: string | null;
+  message: string;
+  title: string;
+}
+
+const FAILURE_STATUSES = new Set<SessionRunStatus>(["cancelled", "expired", "failed"]);
+
+function getFailureFallback(
+  status: SessionRunStatus,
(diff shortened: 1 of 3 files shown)


RockChinQ:
Playwright E2E evidence for the reproduced failed Thread. The persisted `runtime.provision_failed` message is visible inline, and **View process** opens the process dialog.

**Desktop**

Failed Thread error details - desktop

**Mobile (390px)**

Failed Thread error details - mobile

PR #395, opened 2026-07-24
fix(web): unthrottle session stream rendering

Yevanchen:
## Summary

- Remove the per-frame character budget and boundary slicing from the session stream render scheduler: everything queued for the session is delivered on the next frame (requestAnimationFrame batching is the smoothing), with a 512-events-per-frame flood guard for pathological storms.
- Add a 50ms timer fallback alongside requestAnimationFrame so a throttled or paused rAF (hidden/occluded window, battery saver) cannot starve the transcript while the socket keeps receiving events.

## Why

Two production Preview complaints after server TTFT dropped to a few seconds:

1. Visible streaming crawled (~24 chars/frame smooth tier, backward punctuation slicing shrinking CJK slices further), and every slice committed a fresh React state, re-rendering the growing message per frame — the artificial typewriter became the dominant visible latency now that the server is fast.
2. Replies visible in Logs but not in Preview: the queue drains only on rAF; when the browser throttles or pauses frames the transcript freezes despite live socket delivery. The timer fallback keeps output flowing in that mode.

## Verification

- Commands: `bun run --filter @mosoo/web tc`, `bun run --filter @mosoo/web test` (182 pass), `bun run fmt:check:path`.
- Manual steps: pending production deploy verification on try.mosoo.ai Preview (fast gpt-5.5 stream).
- Not run: full `just check` locally; PR CI runs the repository gate.

## Impact

- User/API/contract changes: deliberate UX change — text renders as fast as it arrives instead of a paced typewriter reveal.
- Generated files / GraphQL / DB / lockfile: none.
- Risk and rollback: scheduler-local; revert restores paced rendering. Scheduler host interface gains cancelTimeout/requestTimeout (single consumer updated).

## Review

- Closest review areas: `session-stream-render-scheduler.ts` (drain and dual-arm scheduling), rewritten unit tests incl. the timeout-fallback path.
- Known trade-offs: a very large single delta now commits in one frame; bounded by the flood guard and typical delta sizes.

### DIFF
--- a/apps/web/src/domains/runtime/session-stream/session-stream-render-scheduler.ts
+++ b/apps/web/src/domains/runtime/session-stream/session-stream-render-scheduler.ts
@@ -1,15 +1,15 @@
-import type { AgUiSessionEvent, TextMessageContentEvent } from "@mosoo/ag-ui-session";
+import type { AgUiSessionEvent } from "@mosoo/ag-ui-session";
 
 interface QueuedSessionEvent {
-  enqueuedAt: number;
   event: AgUiSessionEvent;
   sessionId: string;
 }
 
 export interface SessionStreamRenderSchedulerHost {
   cancelFrame: (handle: number) => void;
-  now: () => number;
+  cancelTimeout: (handle: number) => void;
   requestFrame: (callback: () => void) => number;
+  requestTimeout: (callback: () => void, delayMs: number) => number;
 }
 
 export type SessionStreamRenderSchedulerApply = (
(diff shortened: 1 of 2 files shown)


Yevanchen:
CI check note (2026-07-24 10:06 UTC): I reviewed the current checks for this PR. The only non-green item is `Build linux/amd64 and linux/arm64`, which is still pending in https://github.com/langgenius/mosoo/actions/runs/30079761760 at the `Build and publish image` step. The repository check, CodeQL, PR title/commit checks, ship policy, and CLA checks are passing.

GitHub does not expose logs for this in-progress job yet; the direct job-log API returns no log blob, so there is no actionable code failure to patch from this branch at the moment. I did not push changes. If this remains stuck, the next practical action is to cancel/rerun the container workflow.

PR #397, opened 2026-07-24
fix(db): retain usage-event idempotency after detail rollup

### LINKED ISSUE #328: fix(db): retain usage-event idempotency after detail rollup
## Area

Database

## Summary

**Severity: P2 today; P1 if this ledger is later used for quotas, credits, chargeback, or billing.**

The `(source, source_event_id)` uniqueness boundary for model usage exists only on raw `usage_event` rows. Daily rollup aggregates rows older than seven days into `usage_daily_rollup`, then deletes the raw rows and therefore deletes the only durable record of each source identity.

If the runtime later replays or corrects the same logical usage event after its raw row was rolled up, the normal usage upsert sees no conflict and inserts the source identity again. The next rollup adds that full event to the existing aggregate, so one provider/model-call fact is counted twice.

This causes App Usage, CSV exports, model/Agent breakdowns, and later aggregate consumers to overstate consumption. It does not imply that the provider charged twice; the reproduced failure is duplicate accounting of one source event. No remote production database was inspected.

## Steps to reproduce

1. Check out the current #326 stack (`fix/db-ledger-reconciliation`) so the reproduction includes #324, #325, and #326 rather than testing an already-known pre-fix state.
2. Create the local SQLite D1 fixture with the real `usage_event` uniqueness index and `usage_daily_rollup` primary key.
3. Invoke the real `recordRuntimeUsageEvent()` service with:
   - a stable `driverInstanceId`;
   - `nativeCallId: "native-call-1"`;
   - non-zero token usage and a declared usage contract;
   - `createdAtMs` more than seven days before the rollup time.
4. Run the real `runUsageDailyRollup()` service.
5. Confirm the first result:

   ```sql
   SELECT COUNT(*) FROM usage_event;                  -- 0
   SELECT request_count FROM usage_daily_rollup;     -- 1
   ```

6. Replay the exact same `recordRuntimeUsageEvent()` input. Its calculated `(source, source_event_id)` is byte-for-byte identical to the first event.
7. Run `runUsageDailyRollup()` again at the same logical time.
8. Query the aggregate and observe `request_count = 2`; `total_cost_usd_micros` is exactly twice the first aggregate.

## Expected behavior

Retention of raw detail must not shorten the financial idempotency window.

Once a usage source identity has contributed to a daily aggregate, replaying that identity must not add the same tokens or amount again. The implementation needs a durable source receipt, tombstone, or equivalent rollup-aware identity boundary whose retention matches the ledger's correctness requirements.

Legitimate late corrections also need explicit semantics: either apply a proven delta to the aggregate or remain observable and fail closed. They must not be treated as a new full usage fact merely because raw detail expired.

## Actual behavior

The first rollup deleted the only row protected by `usage_event_source_event_idx`. Replaying the identical event then inserted a new raw row without conflict, and the additive rollup counted it again.

The focused local assertions observed:

- after first rollup: raw event count `0`, aggregate request count `1`;
- after identical replay and second rollup: aggregate request count `2`;
- aggregate USD micros after replay: exactly `2x` the original value.

## Evidence

- Runtime source identity is constructed from `driverInstanceId` plus native call ID or Run/call-key fallback: `apps/api/src/modules/cost/application/cost-usage-event.service.ts:168-171` on the #326 stack.
- The usage upsert conflicts only against the raw table's `(source, source_event_id)` index: `apps/api/src/modules/cost/application/cost-usage-event.service.ts:202-226`.
- Rollup adds excluded values to the existing daily aggregate, then deletes all eligible raw rows: `apps/api/src/modules/cost/application/cost-rollup.service.ts:99-123`.
- `usage_daily_rollup` contains aggregate dimensions but no source identity or durable dedupe receipt: `pkgs/db/src/schema/usage.schema.ts`.
- Focused reproduction command:

  ```text
  just test-file apps/api/tests/tmp-cost-rollup-idempotency-repro.test.ts
  1 pass, 0 fail, 4 assertions
  ```

The temporary test exercised the real cost usage service, real rollup service, and SQLite-backed D1 batch. It did not replace either business service with a mock.

I searched current open and closed issues and PRs. #322/#325 make model-call and usage-ledger writes atomic before rollup. #323/#326 conservatively audits historical missing ledger rows and intentionally refuses unsafe old-detail repair. #286/#301 cover aggregate retention length. None retains a live usage source identity after raw detail is deleted or prevents a rolled source event from being counted again.

## Environment

- Verified branch: `fix/db-ledger-reconciliation` (PR #326)
- Verified commit: `fa532e2d31b93b38f5f60f7f0d8ad8145b2a3879`
- Includes prerequisite PRs #324 and #325
- Bun `1.4.0-canary.1`
- Local SQLite D1 fixture using the real service and D1 batch paths
- No remote Cloudflare account, Worker, D1, Queue, provider credential, provider request, or billable resource

This work should follow #324 -> #325 -> #326 because it relies on their forward atomicity, cost-only ingestion, and conservative reconciliation semantics and overlaps the same cost/rollup modules.

## Contribution

I can submit a PR.

## Checklist

vr-varad:

## Summary

- added a new table `usage_event_rollup_receipt`
- sql file for that
- addition of unqiue events in this table
- checking the table for same event before adding `usage_event`
- closes #328 

## Why
- so that same event  ( with same source and source_id) is not charged again, after the earlier copy got cleaned in daily rollup

## Verification

- Commands:
- Manual steps:
- Not run:

## Impact

- User/API/contract changes:
- Env or config changes:
- Risk and rollback:

## Review

- Closest review areas:
- Known trade-offs:


vr-varad:

Yevanchen (changes requested):
Reviewed against #328, including a full verification of the generated `pkgs/db/drizzle` changes. The approach is right — this is the durable source receipt #328 asked for, not a symptom patch. One blocking issue keeps it from merging today: the PR breaks 6 existing tests, which is exactly why the `Repository check` job is red.

## What I verified

**Fix design (matches #328's expected behavior)**

- `usage_event_rollup_receipt` is written in the **same atomic D1 batch** that folds raw rows into `usage_daily_rollup` and deletes them (`cost-rollup.service.ts`), so there is no window where raw detail is gone but the receipt is missing.
- `recordRuntimeUsageEvent` consults the receipt before inserting, and `resolveUsageEventIdentity()` is shared between the check and the insert, so the identity derivation cannot drift. The new early `hasRecordableRuntimeUsage` return is behavior-preserving (the insert builder already returned `null` in that case).
- Receipts are pruned by `rolled_up_at` using the same `DAILY_ROLLUP_RETENTION_DAYS` constant, so a receipt always outlives the aggregate it protects while avoiding unbounded growth.
- `usage_event` has a single ingest path (`recordRuntimeUsageEvent`), so the guard covers all writers, and the rollup records receipts for every source.

**The large generated file (`meta/0003_snapshot.json`, +7,182)**

- After jq normalization it is identical to `0002_snapshot.json` except for the new table and the `id`/`prevId` fields, and `prevId` chains correctly to 0002's id.
- The table definition in the snapshot matches `0003_usage-rollup-receipt.sql` and `usage.schema.ts` exactly.
- `bunx drizzle-kit generate` on this branch reports **"No schema changes, nothing to migrate"** — schema, migration SQL, journal, and snapshot are in sync.
- Appending migration 0003 follows the same pattern as 0001 (#320) and 0002 (#338), per CONTRIBUTING.md's append rule. So the big file is expected: every appended migration commits a full snapshot.

**New tests**

`cost-usage-idempotency.test.ts` reproduces #328's scenario end-to-end with the real services (record → rollup → replay → rollup) and asserts exact totals, plus a receipt-retention pruning case. Good coverage.

## Blocking: 6 existing tests fail

`recordRuntimeUsageEvent` now queries `usage_event_rollup_receipt` on every call, but the fixture in `apps/api/tests/cost-usage-event.test.ts` (`createUsageEventDatabase()`) never creates that table, so all 6 tests in that file fail with:

```
SQLiteError: no such table: usage_event_rollup_receipt
```

The reconciliation fixture got the new table; this one was missed. Adding the same DDL to that fixture fixes it — verified locally, 6/6 pass:

```sql
CREATE TABLE usage_event_rollup_receipt (
  rolled_up_at integer NOT NULL,
  source text NOT NULL,
  source_event_id text NOT NULL,
  PRIMARY KEY (source, source_event_id)
);
```

## Non-blocking notes

- The receipt check and the subsequent `usage_event` insert are two statements, so a replay landing exactly while a rollup batch commits between them could still slip through. Given D1's serialized writes and the daily cadence, that window is negligible and no worse than pre-fix behavior — but if you want to close it completely later, folding the check into the insert (`INSERT … SELECT … WHERE NOT EXISTS`) would make it atomic.
- `_journal.json`'s `when: 1784200000000` and the 0003 snapshot id look hand-authored rather than emitted by drizzle-kit. The chain verifies clean so it's fine here, but per CONTRIBUTING.md (`pkgs/db/drizzle/**` must not be hand-edited), prefer letting `drizzle-kit generate --name <name>` produce these files next time.

Happy to merge as soon as the fixture fix lands and `Repository check` is green.

vr-varad:
@Yevanchen, some dependency issue in CI, otherwise this should be fixed

vr-varad:
hey @Yevanchen any else remaining? or good to review?

Yevanchen (approved):
The blocking issue from my previous review is resolved: commit 3eb5d60 adds the missing `usage_event_rollup_receipt` DDL to the `createUsageEventDatabase()` fixture in `apps/api/tests/cost-usage-event.test.ts`, matching the schema exactly.

Re-verified locally on this branch merged with latest `main`: all 6 previously-failing tests in `cost-usage-event.test.ts` pass, and the full API suite is green (1058 pass / 0 fail). `Repository check` and all other required checks are green on the head commit. Approving and merging.

vr-varad:
Thanks @Yevanchen would love to pick up another issue, if any open for contribution

PR #399, opened 2026-07-24
build(driver): adopt @mosoo/agent-driver package name and bump pin

### LINKED ISSUE #396: chore: bump agent-driver submodule pin past the @mosoo/agent-driver rename
The driver repo renamed its package from `agent-driver` to `@mosoo/agent-driver` (npm publication work, driver #70–79). A naive submodule pin bump now breaks the superproject workspace: `agent-driver@workspace:*` fails to resolve, and every `agent-driver/boot` / `agent-driver/paths` / `agent-driver/orpc` import in apps/api dangles.

Blocked value waiting on this bump (both already merged in the driver repo):
- driver #80 — correct `anthropic_bucketed` cache-token accounting for the acp runtime (fixes an under/over-billing risk in the old contract)
- driver #65 — runtime image −24% uncompressed / −35% gzip

Scope:
1. Update the root workspace dependency and all `agent-driver/*` import specifiers to `@mosoo/agent-driver/*` (or add a package alias if a two-step migration is preferred).
2. Triage the driver test failures observed at head `e9f3e570` when run in-tree (7 fail incl. "ACP file system bridge"; may be workspace-resolution fallout — retest after step 1).
3. Full `just check` + a small staging screen (deploy to a perf stack, few paired pings) before merge — the bump also pulls in the untested #70–79 packaging commits.

Refs: langgenius/mosoo#387, driver PRs #65 #80, review notes on both.

Yevanchen:
Closes #396.

Points the driver submodule at driver main `9d4415d` — first pin bump since the package rename and Driver Contract v2. Brings in:
- langgenius/mosoo-agent-driver#77 — keep a streamed reply on one message when `message_start` is lost (root cause of the production duplicate-bubble fracture, session `01KY9Z5JEVPQP3F548XEK21MEG`)
- langgenius/mosoo-agent-driver#65 — remove npm cache from the runtime image (container-boot diet)
- langgenius/mosoo-agent-driver#80 — acp cache-token buckets under `anthropic_bucketed`

Superproject adaptation:
- all `agent-driver/*` imports → `@mosoo/agent-driver/*` (50 files, incl. boundary-test literals)
- `wrangler.toml` container image path `Dockerfile` → `Containerfile` (renamed upstream)
- `sandbox-layout`: derive control-port count from `DRIVER_CONTROL_PORT_MAX` (v2 dropped `DRIVER_CONTROL_PORT_COUNT`; range unchanged 20000–59999)
- **contract v2 `occurredAt`**: envelope `occurredAt` is now an ISO 8601 string. Wire schema admits strings; `events.ts` parses to epoch ms at the persistence boundary. Without this the old `typeof === "number"` guard silently persisted `null` for every driver event timestamp.
- root fmt/lint scoped to `apps/api apps/web`: pinned driver checkout is styled by the driver repo's own config/CI; root formatting no longer rewrites submodule files on `bun run fmt`

Verification: api tc/test 1052 pass, web tc/test 183 pass, driver tc/build/test green (1 known macOS-only `/var`→`/private/var` symlink case, passes on Linux), root fmt:check/lint/tc/docs:check green.

Staging screen gate (paired warm-ping, openai instrument) running — results will be posted before merge.


Yevanchen:
**Staging screen gate: PASS** (paired warm-ping, openai instrument, stage-a=main `8939fedb` vs stage-b=this branch `95d07bbf`)

- 8/8 pairs valid, 0 run failures
- Median server TTFT: 7,748ms (main) vs **6,387ms (this branch)**; paired median Δ = **−1,087ms**, B faster in 7/8 pairs
- Stage medians (main → branch): context_hydration 796→227ms, driver_turn 159→98ms, prepare_run 4,557→4,086ms

The gate also caught a live incompatibility before merge: the first stage-b deploy failed **every** run (`driver.command_failed.input.start`, oRPC 500). Root cause: Driver Contract v2 emits `runtime.timing.recorded` timestamps as ISO strings (`completedAt`/`startedAt`), while `pkgs/runtime-events` still required legacy epoch-ms fields and threw — poisoning every driver event batch that contained a timing event. Fixed in `95d07bbf` (accept both shapes, normalize to epoch ms, regression test), verified by rerunning the full gate with zero failures. Static checks (tc, CI) could not catch this drift — the two packages type-check independently.

PR #407, opened 2026-07-27
perf(web): avoid quadratic session-stream queue drains

Yevanchen:
## Summary

- Replaced session stream render scheduler front-splice drains with an offset-backed queue and amortized compaction.
- Added regression coverage for flushing after a compaction and rejected-batch requeue so complete event order is preserved.

## Why

- Long streaming sessions are drained across multiple animation frames. Splicing from the front on each frame repeatedly shifted remaining queued events, making large bursts trend toward O(n^2). The offset-backed path keeps normal drains O(n) total while preserving per-session ordering and rejected-batch requeue behavior.

## Verification

- Commands:
  - `just fmt-check-path apps/web/src/domains/runtime/session-stream/session-stream-render-scheduler.ts`
  - `just fmt-check-path apps/web/tests/session-stream-render-scheduler.test.ts`
  - `just test-file apps/web/tests/session-stream-render-scheduler.test.ts`
  - `just test-package @mosoo/web`
  - `just commit-check`
  - `just check`
- Manual steps: N/A
- Not run: N/A

## Impact

- User/API/contract changes: No API or contract changes; streaming UI queue drains avoid repeated array shifts during large event bursts.
- Env or config changes: N/A
- Risk and rollback: Low; localized scheduler queue internals. Roll back this commit if stream batching regresses.

## Review

- Closest review areas: Web session stream render scheduler drain, flush, and rejected-batch requeue behavior.
- Known trade-offs: The static scanner still flags normal loops/map calls in this file; this patch targets the actual repeated front-array shift hotspot.

### DIFF
--- a/apps/web/src/domains/runtime/session-stream/session-stream-render-scheduler.ts
+++ b/apps/web/src/domains/runtime/session-stream/session-stream-render-scheduler.ts
@@ -47,6 +47,7 @@ export class SessionStreamRenderScheduler {
   #frameHandle: number | null = null;
   readonly #host: SessionStreamRenderSchedulerHost;
   #queue: QueuedSessionEvent[] = [];
+  #queueOffset = 0;
   #timeoutHandle: number | null = null;
 
   public constructor(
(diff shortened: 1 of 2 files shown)


Yevanchen:
Superseded by #444 after moving the reviewed commit to the formal `perf/web-session-stream-queue-drain` branch, adding persistent compaction/requeue coverage, and rebasing on current `main`. #444 is the canonical merge path.

PR #433, opened 2026-07-29
feat(web): add two-lane App Overview onboarding

Yevanchen:
## Summary

- Split the pre-deploy App Overview onboarding into explicit coding-agent and console lanes.
- Make the console path match runtime requirements: a provider credential is required, while an API token is optional for external API use.
- Count the run step only after a UI Thread has an actual Run; Preview and unrun Sessions no longer advance progress.
- Preserve the selected lane locally and keep setup actions linked to their owning console pages.

## Verification

- `just test-package @mosoo/web` (194/194)
- Focused formatting checks
- React Doctor (100/100)
- `just commit-check`
- Full workspace check reaches an existing macOS `/var` versus `/private/var` Driver fixture failure reproduced on clean `main`.

## Impact

- Web-console onboarding behavior only; no API, schema, generated-file, or environment changes.
- Query failures still degrade to actionable numbered steps instead of blocking the overview.
- Revert the squash merge to roll back.

### DIFF
--- a/apps/web/src/routes/app-overview/app-overview-install.tsx
+++ b/apps/web/src/routes/app-overview/app-overview-install.tsx
@@ -1,29 +1,64 @@
-import { Check, KeyRound } from "lucide-react";
+import { MousePointerClick, SquareTerminal } from "lucide-react";
 import type { ReactElement } from "react";
 import { useState } from "react";
-import { Link } from "react-router-dom";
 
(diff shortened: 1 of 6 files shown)


Yevanchen:
Superseded by #447 after moving the corrected onboarding state machine to the formal `feat/web-overview-setup-lanes` branch and rebasing it on current `main`. #447 is the canonical merge path.

Issue #446, opened 2026-07-29
feat(runtime): reconcile external tool effects for idempotency-capable MCP providers


AsperforMias:
## Area

Runtime

## Problem

#412 prevents a Driver or control-plane failure from automatically repeating an uncertain write-capable MCP call. Its safe outcome is a durable unknown effect that requires manual user or operator resolution.

That is correct for generic MCP, but it leaves unnecessary manual work for MCP providers that already offer an idempotency-key lookup, a durable operation receipt, or a documented compensating action. Mosoo has no provider-specific reconciliation capability, so it cannot use those provider guarantees to turn an unknown effect into a confirmed result or a confirmed non-execution.

## Reproduction / current limitation

Prerequisite: apply the durable external-effect behavior from #412.

1. Use the local Driver/API fault-injection setup in apps/api/tests/driver-finalization-repair.test.ts with an MCP provider fixture that records a write by the Mosoo idempotency key and exposes a lookup by that key.
2. Execute one mcp.execute command and let the fixture commit the write.
3. Drop the provider response or Driver terminal receipt immediately after the write, then restart the Driver or finalize its control connection.
4. Re-deliver the command.
5. Observe that Mosoo correctly marks the effect unknown and does not invoke the provider again, but it also does not invoke the provider lookup. A user or operator must resolve it manually even though the provider can prove the outcome.

This path is locally reproducible with a fixture; no Cloudflare account, remote MCP server, credentials, or billable provider request is required.

## Proposal

Add an opt-in provider capability for external-effect reconciliation:

- A provider adapter may declare lookup by Mosoo stable idempotency key and return a typed confirmed-success, confirmed-not-executed, or still-unknown result.
- Recovery may reconcile only effects whose provider advertises that capability. It must persist the returned receipt and result before any terminal redelivery.
- Compensation remains a separate explicit per-tool capability. Mosoo must never infer or invent a rollback.
- Providers without the capability retain #412 behavior: unknown stays fenced and no automatic retry occurs.

Keep the adapter contract small, runtime-owned, and provider-specific; do not change the generic MCP wire protocol or treat arbitrary metadata as proof of completion.

## Alternatives considered

- Always retry unknown effects: unsafe because it can duplicate writes.
- Add a universal MCP receipt or compensation protocol in Mosoo: impossible without provider support and would create a private incompatible protocol.
- Keep manual resolution for every provider: safe and remains the fallback, but wastes provider capabilities when they exist.

## Compatibility and migration

[...]

PR #48, opened 2026-06-16
feat(web): replace Debug dropdown with direct Terminal tab in agent editor toolbar

Yevanchen:
## Summary

- Replaces the "Debug" dropdown tab in the agent editor top toolbar with a direct "Terminal" tab. Clicking it switches straight to terminal mode — no dropdown menu in between.

## Why

- The "Debug" dropdown only ever contained a single item ("Terminal"), so the extra dropdown interaction added a needless click. A direct tab is simpler and matches the other toolbar tabs (Dev / Preview / Logs / Cost).

## Verification

- Commands: `tsc --noEmit` on `apps/web` shows no errors for the edited file (only the pre-existing, unrelated `vite-plus/client` generated-types warning).
- Manual steps: open an agent as owner → toolbar now shows a "Terminal" tab (with terminal icon) instead of a "Debug ▾" dropdown; clicking it opens terminal mode and the tab highlights as active.

## Risk And Blast Radius

- Affected packages: `apps/web`
- Affected user flows or APIs: agent editor toolbar navigation only. Terminal access gating (owner-only, `agentKindSupportsOwnerTerminal`) is unchanged — the same `canShowAgentDebugMenuItem` policy still controls whether the entry renders.
- Rollback: revert this commit.

## Evidence

- UI/UX: visible change is the toolbar tab. Before: `Debug ▾` dropdown whose only item is `Terminal`. After: a direct `Terminal` tab.

## Compatibility

- Public contract changes: none
- Env or config changes: none
- Deployment or migration: none

## Reviewer Focus

- Closest review areas: `apps/web/src/routes/agent/agent-detail.route.tsx` — removed `DropdownMenu`/`ChevronDown`/`DEBUG_MODES` usage; render `debugItems` as direct tab buttons.

## Required Checks

- [x] PR title: `type(scope): subject`
- [x] Type: `feat`
- [x] Subject starts lowercase
- [x] Branch follows repo convention
- [x] Commits: `type(scope): subject`, English only
- [x] Diff scoped; no unrelated cleanup
- [x] UI/UX: change described in Evidence
- [x] Generated files: none touched

PR #340, opened 2026-07-16
fix(environment): align package managers with driver runtime

AsperforMias:
Fill what changed. Use N/A for irrelevant or maintainer-only items. See `CONTRIBUTING.md` for branch, CLA, generated file, and CI rules.

## Summary

- Define one shared writable Environment package-manager set (`npm` and `pip`) for Contracts, API validation, and the Web selector.
- Preserve historical `apt`, `cargo`, `gem`, and `go` values for reads while rejecting new revisions and provisioning with an actionable migration error.
- Guard every revision-producing path, including lower-level persistence, forks, and environment-variable updates, so unsupported managers cannot bypass route-level validation.
- Reuse the asynchronous package-artifact build/Backup/Restore path merged in #345 instead of retaining a second runtime-install path.
- Update `apps/driver` to the capability manifest and image verification from [langgenius/mosoo-agent-driver#48](https://github.com/langgenius/mosoo-agent-driver/pull/48), based on Driver #49 and #50.

## Why

- #337 was caused by product write capabilities drifting from what the maintained Driver image could execute. That allowed a valid-looking Environment to fail only when a user Run reached Sandbox provisioning.
- #345 now owns reusable npm/pip package artifacts. This PR completes the product contract: one writable set, explicit legacy read compatibility, persistence-boundary enforcement, and a cross-repository check against the Driver image manifest.
- Depends on [langgenius/mosoo-agent-driver#48](https://github.com/langgenius/mosoo-agent-driver/pull/48). **Merge the Driver PR first. If its merge strategy changes the commit SHA, refresh this PR's `apps/driver` pointer to the resulting upstream commit and rerun `just check` before merging.**

Closes #337.

## Verification

- Commands:
  - `just check`: passed formatting, docs links, lint, typecheck, all workspace tests, and GraphQL freshness; API reported `1022` tests across `172` files with no failures.
  - `just commit-check`: passed; the exact `upstream/main..HEAD` range also passed commit-policy validation.
  - Focused package-manager suites: API `10` passed, Web `3` passed, Contracts `2` passed, Driver artifact contract `6` passed.
  - #345 artifact/build/restore regression suites: `44` passed.
  - Driver #49 child-process environment regression suites: `17` passed.
  - Driver standalone suite: `205` passed and `4` credential-backed live tests skipped; lint, typecheck, and build passed.
  - Current amd64 Driver image build and isolated-prefix smoke passed for `prettier@3.3.3` and `six==1.16.0`.
[...]

PR #449, opened 2026-07-29
fix(runtime): persist external tool effects

### LINKED ISSUE #412: fix(runtime): persist uncertainty for external tool effects
## Area

Runtime

## Summary

**Severity: P0 for write-capable tools.** Mosoo cannot distinguish an external tool action that never started from one that succeeded after the Sandbox or Driver lost its control connection before the terminal receipt was persisted.

The Driver tracks command replay and terminal delivery only in process memory. MCP execution returns an in-memory result, but no durable effect intent, provider receipt, reconciliation record, compensation contract, or stable idempotency key is persisted or forwarded to the tool provider. On terminal Driver loss, the API marks accepted commands and the Run as failed. The recovery code only logs a future requeue decision.

A user resend or future automatic resume can therefore repeat an already successful external write; avoiding replay instead leaves an indeterminate operation incorrectly represented as failed. This affects MCP tools and any Agent tool that can deploy, mutate production data, send messages, or charge money.

## Steps to reproduce

1. Check out main at b4a5f674bcc05ad5ba5221ea1d82c95dbc20e6b9.
2. In the Driver boundary fixture, provide an MCP executor whose tool body increments a local counter and then returns a successful result.
3. Let the Driver send the accepted command update, but make every terminal completed update throw to simulate a control socket loss after the local external effect succeeded.
4. Run the real DriverCommandDispatcher. It retries terminal delivery three times and exits without a persisted completion receipt; the counter is 1.
5. Start a fresh DriverCommandDispatcher with the same MCP command. Its in-memory command history is empty, so the tool body runs again and the counter becomes 2.
6. Separately run the real API finalization path against an accepted command. It terminalizes the command and Run as failed; it does not create an effect record, query a provider receipt, compensate, or enqueue a recovery run.

## Expected behavior

Before a write-capable tool call, Mosoo should durably record an effect intent linked to the Run and command. The effect must have a stable idempotency key, lifecycle state including unknown, attempt history, and an optional external receipt.

For providers that support it, Mosoo must pass the stable idempotency key and persist the returned provider reference before treating the effect as succeeded. After Driver/Sandbox loss, recovery must reconcile unknown effects before any replay. Unsupported or non-idempotent tools must stay unknown and require explicit user or operator resolution.

Compensation should be an explicit per-tool capability, not an assumed rollback. A Saga-like compensating action may run only where the provider declares one.

## Actual behavior

The DriverCommandDelivery dedupe map is process-local. A newly created Driver has no durable record of an already executed external call. The remote MCP executor invokes the tool directly and returns only output text, request id, server id, and tool name.

If the terminal receipt cannot be delivered, the Driver exits. API terminal finalization changes accepted commands to failed and changes the Session Run to failed. The reclaim module records a requeue recommendation only; it does not dispatch it. The stored state has no executed, not-executed, or unknown classification for the external action.

## Evidence

- DriverCommandDelivery stores tracked commands in an in-memory Map: apps/driver/src/core/driver-command-delivery.ts.
- The MCP executor calls the external tool directly and has no Mosoo effect idempotency or receipt persistence: apps/driver/src/runtimes/mcp/remote-http-mcp-executor.ts.
- The local fault injection described above executed the fake external tool once, dropped all three terminal completion attempts, then executed the same command a second time in a fresh Driver. No Cloudflare, MCP credential, or remote system was used.
- `bun --cwd apps/driver test tests/driver-runtime-boundary-state.test.ts` passes, but covers duplicate delivery only while the same Driver remains alive.
- `just test-file apps/api/tests/driver-finalization-repair.test.ts` passes and demonstrates the current terminal behavior: accepted command and Run become failed while only a requeue decision is logged.
- `just test-file apps/api/tests/session-run-reclaim-recovery.test.ts` passes and confirms that recovery is a decision function, not an executed requeue.
- The product contract already says runtime loss is reported and the user must deliberately resend: docs/prd/session-lifecycle.md.

I searched current and closed issues. #329 fixes atomic Session Run admission and is no longer applicable. The merged #220 work classifies reclaim and records a requeue decision, but explicitly defers executing it and does not model external tool effects. #306 and #318 cover Cloudflare deployment resource reconciliation, not generic Sandbox tool effects.

## Environment

- mosoo commit: b4a5f674bcc05ad5ba5221ea1d82c95dbc20e6b9
- Bun: 1.4.0-canary.1
- Local Driver test harness and SQLite D1 API fixture
- No remote Cloudflare account, Sandbox, MCP server, provider credential, model request, or billable resource

## Contribution

I can test a fix.

## Checklist

AsperforMias:
## Summary
- Persist MCP external-effect intent, attempt, result, and unknown-fence state in D1.
- Redeliver a durable MCP result after terminal-receipt loss without invoking the provider again.
- Keep a terminal Driver record while it has an executing or unknown external effect, so routine expiry cleanup cannot cascade-delete the unresolved fence.
- Require Driver protocol v2 and the durable effect ledger for MCP execution.

## Why
- Closes #412: a Driver restart could otherwise replay a write-capable MCP call after an ambiguous terminal delivery.

## Verification
- `just check`
- `just test-file apps/api/tests/driver-instance-record.test.ts`
- `just test-file apps/api/tests/driver-finalization-repair.test.ts`
- Local fault injection uses D1 plus the real Driver Dispatcher to simulate provider success followed by lost terminal delivery, restarts the Dispatcher, and asserts exactly one provider call.
- Not run: live remote MCP or Cloudflare resources (not available locally).

## Impact
- User/API/contract changes: Driver control protocol v2; an unknown effect fails closed with its durable effect ID.
- Generated files / GraphQL / DB / lockfile: additive `0005_external-tool-effects` plus Drizzle snapshot/journal; no GraphQL or lockfile changes.
- Risk and rollback: the migration is additive; old Driver versions are intentionally rejected to prevent unsafe execution.

## Review
- Closest review areas: Driver RPC, API runtime persistence, D1 migrations, Agent Driver MCP dispatcher.
- Known trade-offs: generic MCP has no universal reconciliation or compensation protocol, so unknown effects never auto-replay. Follow-up #446 tracks provider-specific reconciliation.
- Dependency: Driver #89 is merged and provides the runtime persistence protocol. Mosoo main currently relies on the package-manager manifest introduced by its earlier Driver gitlink; Driver #48 restores that manifest on current Driver main. Therefore this PR must not merge before #48, and its repository check remains blocked until that Driver dependency is merged and the gitlink is refreshed to the resulting main commit.

### DIFF
--- a/apps/api/src/modules/sessions/application/session-process-events.service.ts
+++ b/apps/api/src/modules/sessions/application/session-process-events.service.ts
@@ -29,11 +29,14 @@ export interface SessionEventProcessRow {
 
 const DEFAULT_PROCESS_EVENT_LIMIT = 500;
 const MAX_PROCESS_EVENT_LIMIT = 1000;
+// Longer gaps may be permission or user waits, not continuous execution.
+const MAX_INFERRED_PROCESS_EVENT_DURATION_MS = 5 * 60 * 1000;
 
 interface ProcessEventProjection {
   event: SessionProcessEvent;
   endMs: number;
   order: number;
+  runId: SessionRunId | null;
   startMs: number;
 }
 
@@ -63,12 +66,18 @@ function finalizeProcessEventDurations(
 
   return sortedProjections.map((projection, index) => {
     const next = sortedProjections[index + 1] ?? null;
+    const inferredDurationMs = next === null ? 0 : next.startMs - projection.startMs;
+    const canInferDuration =
+      projection.runId !== null &&
+      next?.runId === projection.runId &&
+      inferredDurationMs >= 0 &&
+      inferredDurationMs <= MAX_INFERRED_PROCESS_EVENT_DURATION_MS;
     const durationMs =
       projection.endMs > projection.startMs
         ? projection.endMs - projection.startMs
-        : next === null
-          ? 0
-          : Math.max(0, next.startMs - projection.startMs);
+        : canInferDuration
+          ? inferredDurationMs
+          : 0;
 
     return {
       content: projection.event.content,
@@ -97,6 +106,7 @@ function toProcessEventProjectionFromSessionEventRow(
       type: row.process_type,
     },
     order: row.seq,
+    runId: row.run_id,
     startMs: row.occurred_at,
   };
 }
--- a/apps/api/tests/session-process-events.test.ts
+++ b/apps/api/tests/session-process-events.test.ts
@@ -284,6 +284,139 @@ describe("session process event projection", () => {
     expect(events.map((event) => event.type)).toEqual(["run.started"]);
   });
 
+  test("keeps valid explicit durations authoritative", async () => {
+    const database = createProcessEventQueryDatabase();
+    await insertSessionProcessEvent(database, {
+      endedAt: 901_000,
+      id: "event-1",
+      occurredAt: 1_000,
+      runId: "run-1",
+      seq: 1,
+    });
+    await insertSessionProcessEvent(database, {
+      id: "event-2",
+      occurredAt: 2_000,
+      runId: "run-2",
+      seq: 2,
+    });
+
+    const events = await getThreadSessionProcessEvents(database, VIEWER, {
+      appId: APP_ID,
+      sessionId: SESSION_ID,
+    });
+
+    expect(events[0]?.durationMs).toBe(900_000);
+  });
+
+  test("infers adjacent event durations within the same run", async () => {
+    const database = createProcessEventQueryDatabase();
+    await insertSessionProcessEvent(database, {
+      id: "event-1",
+      occurredAt: 1_000,
+      runId: "run-1",
+      seq: 1,
+    });
+    await insertSessionProcessEvent(database, {
+      id: "event-2",
+      occurredAt: 2_500,
+      runId: "run-1",
+      seq: 2,
+    });
+
+    const events = await getThreadSessionProcessEvents(database, VIEWER, {
+      appId: APP_ID,
+      sessionId: SESSION_ID,
+    });
+
+    expect(events.map((event) => event.durationMs)).toEqual([1_500, 0]);
+  });
+
+  test("does not infer durations without an identified shared run", async () => {
+    const database = createProcessEventQueryDatabase();
+    await insertSessionProcessEvent(database, {
+      id: "event-1",
+      occurredAt: 1_000,
+      runId: null,
+      seq: 1,
+    });
+    await insertSessionProcessEvent(database, {
+      id: "event-2",
+      occurredAt: 2_000,
+      runId: "run-1",
+      seq: 2,
+    });
+
+    const events = await getThreadSessionProcessEvents(database, VIEWER, {
+      appId: APP_ID,
+      sessionId: SESSION_ID,
+    });
+
+    expect(events.map((event) => event.durationMs)).toEqual([0, 0]);
+  });
+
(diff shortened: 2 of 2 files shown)


AsperforMias:
这个要driver那边的#89先合 @Yevanchen

Yevanchen (changes requested):
The durable fence is directionally correct, but one persistence invariant is currently lost: unknown effects cascade-delete with their Driver, and maintenance deletes stopped/failed Drivers after 24 hours. That erases the evidence the PR says requires explicit resolution. Please decouple unresolved-effect retention or exclude such Drivers from cleanup and add the retention regression test.

Separately, the PR currently conflicts on apps/driver and points at an obsolete Driver commit; after Driver #89 lands, rebase/update the gitlink. Both this PR and #481 currently allocate migration 0005, so whichever lands second must generate the next append-only migration rather than reusing 0005.

pkgs/db/src/schema/runtime.schema.ts
```diff
@@ -219,6 +226,74 @@ export const driverCommandsTable = sqliteTable(
...
+  "external_tool_effect",
+  {
+    attemptCount: integer("attempt_count").notNull().default(0),
+    commandId: platformIdColumn<DriverCommandId>("command_id")
+      .notNull()
+      .references(() => driverCommandsTable.id, { onDelete: "cascade" }),
+    createdAt: integer("created_at").notNull(),
+    driverInstanceId: platformIdColumn<DriverInstanceId>("driver_instance_id")
+      .notNull()
+      .references(() => driverInstancesTable.id, { onDelete: "cascade" }),
```
Yevanchen, line 248: cleanupDriverInstances() deletes stopped/failed Driver rows at expiresAt (24 hours), and this cascade then deletes unknown effects plus their attempt/receipt evidence. That contradicts the explicit-resolution fence: after one day Mosoo no longer knows the action was uncertain. Preserve unresolved effects independently, or prevent Driver cleanup while an executing/unknown effect exists, and cover that retention boundary.

AsperforMias:
Addressed the requested retention invariant and rebased onto current main. Terminal Driver cleanup now skips records with executing or unknown external effects, preventing the Driver/command cascade from erasing the durable fence. Added direct maintenance and end-to-end persistence regressions; `just check` and both focused tests pass. The submodule now points at merged Driver #89 (`ddbb4305`).

AsperforMias:
CI diagnosis: #449 itself does not fail a runtime-persistence test. Its updated Driver #89 gitlink lacks the package-manager manifest that current Mosoo main expects from the earlier #340 gitlink; the contract test fails with ENOENT for apps/driver/environment-package-managers.json. Driver #48 restores that manifest on current Driver main and is now the required merge-order dependency. I have corrected the PR description accordingly; #449 must be refreshed to the post-#48 Driver main commit before merge.

AsperforMias:
#449 的CI 依赖 driver #48，driver #89 的当前 main 缺少 mosoo #340 所依赖的 `package-manager manifest`。driver #48 合并后需刷新 gitlink，才能合 #449

Yevanchen:
Review update: the retention fix and regression now address the inline durable-fence finding, but the PR is still not merge-ready. Driver #48 merged on Aug 6; please rebase onto current Mosoo `main`, update the Driver gitlink to the post-#48 commit, and generate the next append-only migration with `just db-generate` instead of retaining stale `0005`. Then resolve the remaining review state and rerun `just check`. Issue #412 stays open until this lands.

PR #471, opened 2026-07-30
fix(domain): route assets through redirect worker

Yevanchen:
## Summary

- route static assets through the existing Web Worker so legacy-domain and HTTP redirects also cover `/` and asset-backed SPA paths
- reuse one propagation-aware redirect assertion for legacy deep links, legacy root, canonical HTTP root, and canonical HTTP API
- lock the Worker-first asset setting in the production-origin contract test

## Root cause

Cloudflare Workers Assets serves matching assets before the User Worker by default. That let `https://try.mosoo.ai/` and `http://cloud.mosoo.ai/` return cached HTML 200 while unknown deep links correctly reached the redirect code. `run_worker_first = true` applies the existing redirect policy consistently.

## Verification

- 6 focused API/Web/origin tests passed
- `just fmt-check`
- `git diff --check`
- Web build
- Wrangler production Web dry-run

### DIFF
--- a/.github/workflows/deploy-try.yml
+++ b/.github/workflows/deploy-try.yml
@@ -108,22 +108,30 @@ jobs:
             -H 'content-type: application/json' \
             --data '{"query":"query { __typename }"}' \
             | jq -e '.data.__typename == "Query"'
-          expected_redirect='308 https://cloud.mosoo.ai/domain-migration-check?source=deploy'
-          for attempt in {1..12}; do
-            actual_redirect="$(curl "${curl_args[@]}" --output /dev/null \
-              --write-out '%{http_code} %{redirect_url}' \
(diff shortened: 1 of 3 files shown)


Yevanchen:
CI follow-up for `deploy-try` run https://github.com/langgenius/mosoo/actions/runs/30550497690:

The repo checks, migration checks, artifact build, and API/Web worker dry-runs passed. The failing step is `Deploy`, where Cloudflare rejected the container rollout for `mosoo-api-prod-sandbox-prod`:

```text
Surpassed total account limits: This application requires more than the allocated quota of memory in this account to run all its requested instances of count 1000.
```

I did not push a branch fix here because this PR is already merged, so there is no open original PR branch to update for this failed post-merge deploy run. This looks like it needs either a follow-up change reducing the production container memory/instance request or a Cloudflare quota adjustment.

PR #472, opened 2026-07-30
fix(web): route root assets through worker

Yevanchen:
## Summary
- route only `/` through the Web Worker before static asset serving
- preserve asset-first behavior for every other path
- lock the routing contract with a regression test

## Why
Production deploy run 30549463021 deployed successfully but failed verification because `http://cloud.mosoo.ai/` returned 200. Cloudflare static assets serve matching assets before Worker code unless `assets.run_worker_first` selects the path.

## Verification
- `vp exec bun test apps/web/tests/worker-domain-redirect.test.ts`
- production Web build
- `wrangler deploy --env prod --dry-run`
- `TMPDIR=/private/tmp just check`

### DIFF
--- a/apps/web/tests/worker-domain-redirect.test.ts
+++ b/apps/web/tests/worker-domain-redirect.test.ts
@@ -1,6 +1,13 @@
 import { expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
 
 import worker from "../src/worker";
+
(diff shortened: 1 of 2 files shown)


Yevanchen:
Superseded by #471, which merged the same Cloudflare asset-first root-cause fix and is already on deploy/try. Closing to avoid a conflicting duplicate.

PR #481, opened 2026-08-01
feat(deployment): persist App secrets across redeploys

### LINKED ISSUE #429: feat(deployment): manage persistent App secrets across redeploys
## Area

App Deployment / Secrets

## Problem

A deployed App can need a server-side credential for a third-party or Mosoo Public API, but Mosoo Deployment currently has no App-secret surface. The manifest can bind Agents as generated plain-text capability URLs, while secret values must be added afterward with provider-specific tooling.

This breaks the Deployment abstraction and makes a working App fragile across redeploys.

## Reproduction

The public `Yevanchen/mosoo-codex-pet` App uses Mosoo attachment upload, asynchronous Threads, and artifact download. These Public API routes require an owner PAT kept only in the Worker.

1. Deploy through `mosoo console apps deploy-app`.
2. The Deployment succeeds and binds `MOSOO_AGENT_CODEX_PET_URL`.
3. `POST /api/pets` returns `The Mosoo backend is not configured` because `MOSOO_API_TOKEN` is absent.
4. `wrangler secret put MOSOO_API_TOKEN --name <managed-worker>` fixes the live App; a safe multipart smoke then reaches Mosoo and returns the expected `file is required` validation.

Current Worker upload metadata is built only from Mosoo-resolved plain-text bindings. Cloudflare Worker versions capture bindings as part of version state, so an out-of-band secret is not a durable Mosoo Deployment contract and may be dropped by a later managed redeploy.

## User story

As an App Owner, I want to declare required secret names and set their values in Mosoo so deployments fail fast when a secret is missing and preserve configured secrets across every redeploy without exposing values to source code, build logs, Runs, or API responses.

## Acceptance criteria

- [ ] An App Owner can create/update/revoke an App-scoped secret through Console and CLI/API.
- [ ] Secret values are write-only; list/read surfaces return names and metadata, never values.
- [ ] `.mosoo.toml` can declare required secret names without containing values.
- [ ] Deployment validates required secrets before external submission and returns a stable actionable error when one is missing.
- [ ] Worker uploads bind configured values as Cloudflare secret bindings and preserve them across new versions/redeploys.
- [ ] Deleting the App deployment removes/revokes its managed Worker secrets according to a documented lifecycle.
- [ ] Build Sandboxes, generated Wrangler config, logs, Run payloads, GraphQL, and CLI output never contain secret values.
- [ ] Existing Agent bindings and Apps with no declared secrets remain backward compatible.
- [ ] A regression test covers first deploy, secret rotation, redeploy, and deletion.

## Out of scope

- A general-purpose vault UI
- Client-visible environment variables
- Cross-App shared secrets
- Automatically minting owner PATs without explicit consent

## Workaround

Run `wrangler secret put` against the managed Worker after each Mosoo deployment. This is functional today but should not be the product contract.

## Checklist

- [x] I searched existing open and closed issues before opening this request.
- [x] I reproduced the gap against a successful Mosoo Deployment.
- [x] No secret value is included in this issue.

AsperforMias:

## Summary

- Adds App-scoped deployment secret storage/list/rotate/delete, validates `[secrets].required`, injects Worker `secret_text`, and removes all stored App deployment secrets only after managed Cloudflare cleanup succeeds.
- Checks only required secret names before a potentially long build; after the build, reloads the active deployment context and resolves the current values immediately before Worker submission.
- Makes single-secret deletion explicitly future-deployment-only in the Console and PRD: it removes Mosoo storage but does not claim to remove a binding from an already deployed Worker.
- Adds Console terminal guidance aligned with the `mosoo-connector` / Lathe generated `mosoo console` CLI surface, using JSON body files so secret values stay out of shell history.

## Why

- Fixes #429 by making App deployment secrets persistent across redeploys instead of relying on out-of-band `wrangler secret` state.
- Prevents a value rotated or deleted during a build from being submitted after the build completes.
- Keeps secret values write-only across Console, GraphQL, generated CLI docs, deployment plans, generated Wrangler config, build sandbox state, logs, and API responses.

## Verification

- `just db-reset-local` (fresh local D1 chain including `0005_app-deployment-secrets`)
- `just test-file apps/api/tests/app-deployment-service.test.ts` (30 pass; includes rotation/deletion during build)
- `just test-file apps/web/tests/deployment-secrets-cli.test.ts` (2 pass)
- `just fmt-check-path` for the changed API, Web, test, and PRD files
- Not run: production deploy or live Cloudflare deployment; the PR Repository check is running.

## Impact

- User/API/contract changes:
  - Adds App deployment secret GraphQL fields/mutations and Console UI for write-only App deployment secrets.
  - Adds `.mosoo.toml [secrets].required` deployment contract for Worker deployments.
  - Sends declared values as Cloudflare Worker `secret_text` bindings during deployment.
  - GraphQL generated schema/web outputs are included.
  - Adds append-only D1 migration `0005_app-deployment-secrets` plus Drizzle snapshot/journal updates.
  - No lockfile change.
- Env or config changes:
  - Uses existing `VAULT_ROOT_SECRET`; no new env var or config secret.
- Risk and rollback:
  - Additive API and migration. Roll back behavior by hiding the Console/CLI entry points and not declaring `[secrets].required` in deployment manifests.
  - The migration is append-only and should not be rewritten after merge/application.

## Review

- Closest review areas:
  - App deployment secret lifecycle, Vault integration, Cloudflare Worker upload metadata, GraphQL generated CLI contract, migration/indexes, and deletion ordering.
- Known trade-offs:
  - Installed CLI users need the corresponding `mosoo-connector` / Lathe generated CLI release after this GraphQL contract lands.
  - Deleting an individual stored secret takes effect at the next successful deployment; it is not an immediate live Worker binding deletion.
  - #449 also creates migration `0005`; whichever PR merges second must rebase and generate the next append-only migration.


Yevanchen (changes requested):
Requesting changes for two secret-lifecycle gaps. Secret plaintext is resolved before the potentially long build, so a value rotated or revoked during build can still be submitted afterward. Also, the advertised revoke operation deletes only Mosoo's saved Vault value; an already deployed Worker keeps its existing secret_text binding.

Please resolve values at the final submission boundary and either remove the live binding on revoke or make the API/UI/PRD explicitly describe next-deployment-only deletion. CI is green, but these races are not covered.

apps/api/src/modules/apps/application/app-deployment-executor.service.ts
```diff
@@ -912,10 +919,20 @@ export async function dispatchAppDeploymentRun(
     }
 
     let envVars: Record<string, string>;
+    let secretVars: Record<string, string>;
     try {
-      envVars = await resolveDeploymentEnvVars(bindings, context.deployment, context.run, plan);
+      [envVars, secretVars] = await Promise.all([
```
Yevanchen: Resolve plaintext secret values after runner.build, immediately before runner.deploy and after reloading the current context. As written, a long build retains plaintext and can deploy a value that the owner rotated or revoked while the build was running. A name-only preflight before build is fine, but the actual Vault read needs to happen at the submission boundary.

apps/api/src/modules/apps/application/app-deployment-secret.service.ts
```diff
@@ -0,0 +1,287 @@
...
+  const created = await readAppDeploymentSecretRow(bindings.DB, input.appId, name);
+
+  if (created === null) {
+    throw new Error("App deployment secret could not be loaded after creation.");
+  }
+
+  return toAppDeploymentSecret(created);
+}
+
+export async function deleteAppDeploymentSecret(
```
Yevanchen, line 185: This revoke path removes only the saved Vault/mapping row. A live Worker keeps the previously uploaded secret_text binding until a later redeploy or deployment deletion, so the owner is told a credential was revoked while the deployed app can still use it. Either remove/update the live binding before returning, or rename/document this as deletion for future deployments and warn the user explicitly.

AsperforMias:
Addressed both secret-lifecycle review gaps and rebased onto current main. The build phase now performs name-only preflight; after the build and a current-context reload, the submission boundary resolves the latest vault values. New regressions prove that a rotation during build submits the new value and a deletion during build blocks submission. The Console and PRD now state that single-secret deletion affects future deployments only and does not immediately remove an existing Worker binding. Local D1 migration reset, focused API/Web tests, and API/Web type checks pass.

Yevanchen:
Closing as superseded by #504, which merged the same non-migration implementation and closed #429. The merged version correctly appended `0007_app-deployment-secrets` on the current migration chain; merging this stale branch would duplicate the feature and conflict with immutable migration history.

PR #533, opened 2026-08-13
fix(api): stabilize streamed tool identity

Yevanchen:
## What changed

- Persist the stable tool name only while an ACP tool call is running.
- Persist canonical tool input only once the tool call is terminal.
- Keep terminal display titles out of durable tool identity.
- Add an isolated staging environment that mirrors production topology and capacity.
- Include the runtime-health commit currently deployed in production so the reviewed baseline matches the live API.

## Root cause

Migration `0008_public-thread-tool-call-identity.sql` correctly rejects reused tool-call identities. ACP providers legitimately stream evolving tool arguments, including valid but incomplete JSON objects. Persisting each running snapshot made later snapshots look like identity reuse, so D1 raised `session_event tool identity conflict` and the run surfaced `acp.turn_failed`.

## Impact

Skill loading and subsequent tools such as `bash` can complete without violating the durable identity constraint. Public Thread events expose input only after the terminal snapshot is authoritative.

## Validation

- Production-shaped stage uses isolated D1, R2, queues, Durable Objects, secrets, and the same `standard-2` container image/capacity as production.
- Exact live API base commit plus this hotfix completed a real DeepSeek `acp-fallback` run using the reported Skill path (`01KZY5QECD57Y74XD3H34N8F79`).
- `TMPDIR=/private/tmp just check` passed.
- Production D1 has no pending migrations; required queues exist.
- Driver and Web builds passed.
- API and Web Wrangler production dry-runs passed.

No production resource was changed by the validation steps.


Yevanchen:
Superseded by #534 because the repository ship policy requires a Conventional Commit-style branch name.

PR #627, opened 2026-09-11
fix(runtime): support subpaths and preserve target routing in MCP proxy

### LINKED ISSUE #626: fix(runtime): MCP proxy router drops downstream subpaths and returns 404
### Area
Runtime

### Summary
`driver.all("/mcp/proxy/:serverId", ...)` in `apps/api/src/adapters/http/routes/driver-route.ts` only matches the single exact path segment without wildcard subpath matching (`/*`). When a client or driver sends requests with downstream subpaths (e.g. `/mcp/proxy/:serverId/messages` or `/mcp/proxy/:serverId/sse`), Hono returns 404 Route Not Found. Additionally, `toUpstreamProxyUrl` only copies search parameters and completely ignores incoming request pathnames, dropping any subpaths even if received.

### Steps to reproduce
1. Configure an MCP server with target URL (e.g. `https://mcp.example.com/base`).
2. Mint a runtime action grant for `mcp_proxy`.
3. Send an HTTP request to `/api/driver/mcp/proxy/<serverId>/messages` with the authorization grant.
4. Inspect the HTTP response status.

### Expected behavior
Requests with subpaths should match the proxy route, validate paths against directory traversal, and forward the request to `https://mcp.example.com/base/messages` with original query parameters (excluding the internal `grant` token).

### Actual behavior
1. Hono returns 404 Route Not Found because the router is defined as `driver.all("/mcp/proxy/:serverId", ...)` without wildcard support.
2. `toUpstreamProxyUrl` unconditionally discards incoming subpaths, resolving only `new URL(upstreamUrl)` with search parameters.

### Evidence
- In `apps/api/src/adapters/http/routes/driver-route.ts`, line 783 registers only the single-segment route:
  ```ts
  driver.all("/mcp/proxy/:serverId", async (c) => { ... });
  ```
  In contrast, LLM proxy (line 652) registers `/llm/proxy/:credentialId/*` and extracts subpaths via `extractLlmProxySubPath`.
- `toUpstreamProxyUrl` ignores `request.url` pathname completely:
  ```ts
  function toUpstreamProxyUrl(request: Request, upstreamUrl: string): string {
    const target = new URL(upstreamUrl);
    const incoming = new URL(request.url);
    for (const [key, value] of incoming.searchParams) {
      if (key !== "grant") {
        target.searchParams.append(key, value);
      }
    }
    return target.toString();
  }
  ```

### Environment
- Repository: `langgenius/mosoo`
- Revision: HEAD (`main`)
- Component: `apps/api` (Driver HTTP routes)

### Contribution
I can submit a PR

xsvm:
## Summary

- Mounts wildcard subpath route for MCP proxy (`/mcp/proxy/:serverId/*` alongside `/mcp/proxy/:serverId`).
- Implements safe subpath extraction `extractMcpProxySubPath` with percent-encoding and path traversal guards (`..`, `\`, `%25`, `%2f`, `%5c`, and malformed encoding).
- Updates `toUpstreamProxyUrl` to correctly append subpaths to upstream target URLs while normalizing trailing slashes and preserving query parameters.
- Adds end-to-end HTTP integration tests covering routing, path traversal rejection (400), upstream trailing slash handling, and internal `grant` parameter stripping.
- Fixes #626.

## Why

- Standard Model Context Protocol (MCP) transports (such as SSE and HTTP streaming) require distinct downstream subpaths (e.g. `/sse`, `/messages`, `/tools/call`).
- Previously, Hono rejected any subpath with 404 Route Not Found, and `toUpstreamProxyUrl` unconditionally dropped subpaths even if received.

## Verification

- Commands:
  - `bun test apps/api/tests/driver-mcp-proxy-route.test.ts` (5 pass, 0 fail, 17 expects)
  - `bun test apps/api/tests/driver-llm-proxy-route.test.ts` (33 pass, 0 fail)
  - `bun x tsc -p apps/api/tsconfig.json --noEmit` (0 errors)
  - `vp fmt --check` (100% matched)
  - `bun scripts/validate-commit-range.ts upstream/main HEAD` (1 commit passed)
- Manual steps: N/A
- Not run: N/A

## Impact

- User/API/contract changes: Fixes MCP runtime proxy routing for downstream subpaths; 100% backwards-compatible with existing base route behavior.
- Env or config changes: N/A
- Risk and rollback: Zero regression risk. Clean single-commit revert if needed.

## Review

- Closest review areas: `apps/api/src/adapters/http/routes/driver-route.ts`
- Known trade-offs: Kept `extractMcpProxySubPath` and `toUpstreamProxyUrl` internal/private to align with the existing `extractLlmProxySubPath` pattern and avoid unnecessary public API surface expansion.

## Design (UI changes only, otherwise N/A)

- N/A

### DIFF
--- a/apps/api/src/adapters/http/routes/driver-route.ts
+++ b/apps/api/src/adapters/http/routes/driver-route.ts
@@ -64,6 +64,8 @@ const HOP_BY_HOP_HEADERS = new Set([
 const LLM_PROXY_GRANT_HEADERS = new Set(["authorization", "x-api-key", "x-goog-api-key"]);
 const LLM_PROXY_PATH_MARKER = "/llm/proxy/";
 const LLM_PROXY_UNSAFE_PATH_ENCODING = /%(?:25|2f|5c)/iu;
+const MCP_PROXY_PATH_MARKER = "/mcp/proxy/";
+const MCP_PROXY_UNSAFE_PATH_ENCODING = /%(?:25|2f|5c)/iu;
 const OPENAI_IMAGE_API_PATHS = new Set(["/images/edits", "/images/generations"]);
 
 async function requireDriverActionGrant(c: Context<ApiGatewayEnvironment>) {
@@ -121,9 +123,14 @@ function copyProxyResponseHeaders(headers: Headers): Headers {
   return nextHeaders;
 }
 
-function toUpstreamProxyUrl(request: Request, upstreamUrl: string): string {
+function toUpstreamProxyUrl(request: Request, upstreamUrl: string, subPath = ""): string {
   const target = new URL(upstreamUrl);
   const incoming = new URL(request.url);
+
+  const trimmedSubPath = subPath.replace(/^\/+|\/+$/g, "");
+  if (trimmedSubPath.length > 0) {
+    target.pathname = target.pathname.replace(/\/+$/, "") + `/${trimmedSubPath}`;
+  }
 
   for (const [key, value] of incoming.searchParams) {
     if (key !== "grant") {
@@ -214,6 +221,44 @@ function extractLlmProxySubPath(pathname: string): string | null {
   const subPath = slashIndex === -1 ? "" : rest.slice(slashIndex);
 
   if (subPath.includes("\\") || LLM_PROXY_UNSAFE_PATH_ENCODING.test(subPath)) {
+    return null;
+  }
+
+  for (const segment of subPath.split("/")) {
+    let decoded: string;
+
+    try {
+      decoded = decodeURIComponent(segment);
+    } catch {
+      return null;
+    }
+
+    if (
+      decoded === "." ||
+      decoded === ".." ||
+      decoded.includes("/") ||
+      decoded.includes("\\") ||
+      decoded.includes("%")
+    ) {
+      return null;
+    }
+  }
+
+  return subPath;
+}
+
+function extractMcpProxySubPath(pathname: string): string | null {
+  const markerIndex = pathname.indexOf(MCP_PROXY_PATH_MARKER);
+
+  if (markerIndex === -1) {
+    return null;
+  }
+
+  const rest = pathname.slice(markerIndex + MCP_PROXY_PATH_MARKER.length);
+  const slashIndex = rest.indexOf("/");
+  const subPath = slashIndex === -1 ? "" : rest.slice(slashIndex);
+
+  if (subPath.includes("\\") || MCP_PROXY_UNSAFE_PATH_ENCODING.test(subPath)) {
     return null;
   }
 
@@ -453,6 +498,7 @@ async function proxyRuntimeMcpRequest(
     upstreamAccessToken: string;
     url: string;
   },
+  subPath: string,
 ): Promise<Response> {
   const init: RequestInit = {
     headers: copyProxyRequestHeaders(
@@ -467,7 +513,7 @@ async function proxyRuntimeMcpRequest(
     init.body = request.body;
   }
 
-  const response = await fetch(toUpstreamProxyUrl(request, input.url), init);
+  const response = await fetch(toUpstreamProxyUrl(request, input.url, subPath), init);
 
   return new Response(response.body, {
     headers: copyProxyResponseHeaders(response.headers),
@@ -780,7 +826,7 @@ export function registerDriverRoute(app: Hono<ApiGatewayEnvironment>) {
     }
   });
 
-  driver.all("/mcp/proxy/:serverId", async (c) => {
+  const handleRuntimeMcpProxy = async (c: Context<ApiGatewayEnvironment>) => {
     await cleanupDriverInstances(c.env);
 
     let grant: Awaited<ReturnType<typeof requireDriverAuthorizationGrant>>;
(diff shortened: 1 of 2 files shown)


xsvm:

xsvm:
recheck

apps/api/src/adapters/http/routes/driver-route.ts
```diff
@@ -857,7 +909,10 @@ export function registerDriverRoute(app: Hono<ApiGatewayEnvironment>) {
       const details = toRuntimeMcpProxyPublicErrorDetails(proxyError);
       return Response.json(runtimeMcpProxyErrorBody(details), { status: details.status });
     }
-  });
+  };
+
+  driver.all("/mcp/proxy/:serverId", handleRuntimeMcpProxy);
+  driver.all("/mcp/proxy/:serverId/*", handleRuntimeMcpProxy);
```
Yevanchen, line 915: **[P1] Keep credential-bearing requests within admitted MCP transport targets**

This wildcard accepts every HTTP method and arbitrary descendants, while `resolveRuntimeMcpProxyTarget` receives no subpath to authorize and `proxyRuntimeMcpRequest` injects the stored upstream bearer credential. With a registered URL ending in `/base`, a valid runtime grant now permits `DELETE /api/driver/mcp/proxy/<id>/admin/config` to become an authenticated `DELETE /base/admin/config`. I reproduced the boundary change against a local HTTP fixture: the parent returns 404 without contacting upstream; this commit forwards the credential and returns 200. If the configured prefix contains privileged non-MCP endpoints, or the registered URL is the origin root, Sandbox code gains access to those operations using the owner's credential. Keep Streamable HTTP bound to the configured endpoint; any additional legacy/custom transport targets need an explicit validated mapping and method policy. Please add a denial test proving unrelated descendants cannot receive the upstream credential.

apps/api/src/adapters/http/routes/driver-route.ts
```diff
@@ -121,10 +123,15 @@ function copyProxyResponseHeaders(headers: Headers): Headers {
...
 }
 
-function toUpstreamProxyUrl(request: Request, upstreamUrl: string): string {
+function toUpstreamProxyUrl(request: Request, upstreamUrl: string, subPath = ""): string {
   const target = new URL(upstreamUrl);
   const incoming = new URL(request.url);
 
+  const trimmedSubPath = subPath.replace(/^\/+|\/+$/g, "");
+  if (trimmedSubPath.length > 0) {
+    target.pathname = target.pathname.replace(/\/+$/, "") + `/${trimmedSubPath}`;
```
Yevanchen, line 132: **[P2] Preserve the downstream endpoint's trailing slash**

`subPath.replace(/^\/+|\/+$/g, "")` removes a significant trailing slash as well as the join separator. With upstream `/base`, a request to `/api/driver/mcp/proxy/<id>/messages/` is forwarded to `/base/messages`, although `/base/messages/` and `/base/messages` can be different endpoints. A local upstream accepting only `/base/messages/` returns 200 directly and 404 through this PR. Normalize the base/suffix join without removing the validated suffix's trailing slash, including a suffix of `/`, and add coverage for that case.

Yevanchen (changes requested):
## Review: MCP proxy subpaths at `18de54f`

### Passed

- `just test-file apps/api/tests/driver-mcp-proxy-route.test.ts`: 5 passed.
- `just test-file apps/api/tests/driver-llm-proxy-route.test.ts`: 33 passed.
- `just tc-package @mosoo/api`: passed.
- `just fmt-check-path apps/api/tests/driver-mcp-proxy-route.test.ts`: passed.
- A local HTTP fixture exercising the actual Driver `executeRemoteHttpMcpCommand`, pinned MCP SDK, grant verifier, and Hono route successfully initialized and executed `tools/call` on both the parent route and this PR. All requests used the same proxy URL and the same configured upstream MCP endpoint.

### Findings

1. **[P1] The wildcard expands use of the stored upstream credential to arbitrary descendant paths and methods.** See the inline comment at the new route registration.
2. **[P2] Subpath normalization removes significant trailing slashes and changes the target endpoint.** See the inline comment at URL construction.

### Verification gaps and issue scope

The transport premise in the PR description needs correction. [Streamable HTTP](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#streamable-http) uses one MCP endpoint; `tools/call` is a JSON-RPC method. The checked-in Driver custom executor explicitly uses `StreamableHTTPClientTransport`, and the other backend configurations receive the complete `proxyUrl`.

[Legacy HTTP+SSE](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse) does use a separately advertised POST URI. That compatibility gap remains after this change: with upstream `/base/sse` sending `event: endpoint\ndata: /messages?sessionId=review-session`, the pinned `SSEClientTransport` sends its next POST to `/messages?sessionId=review-session` on the API origin, receives 404, and never reaches the wildcard proxy. This also fails on the parent. The new tests manually choose proxy subpaths and return JSON for the supposed SSE response, so they do not exercise this negotiation.

The local reproduction uses fixture target/vault resolution and a local upstream. It establishes routing behavior, not full Sandbox/provider or production acceptance. I did not run the full `just check` gate. GitHub's [PR Check run](https://github.com/langgenius/mosoo/actions/runs/34596309076) is currently `action_required`, so the metadata checks alone do not establish full CI success.

### Verdict

**NEEDS CHANGES.** Please preserve the credential boundary and exact endpoint path, and provide the actual failing MCP client/server/transport trace from #626 before settling the extra routing contract.

